PR #39, opened 2018-04-30
[WIP] Utilize typeguard components to generic type validation.

asford:
Converts type validation to use logic derived from [typeguard](https://github.com/agronholm/typeguard), rather than writing custom validators for custom types.

Currently in progress, need to profile runtime cost of typeguard components. May need to lift typeguard checks into more specialized converter/validator classes and combine with declaration-time validator/converter instantiation to effectively integration with attrs.


jflat06:
Closing this to clean things up.

PR #159, opened 2019-01-28
Add OMP and AABB cpu dispatch

willsheffler:
Hey Alex, lemme know where you want me to go with this.... just an initial stab at what we talked about. 

The AABB version seems ok... for ubq it's about a 3x speedup over "naive", which seems about right. Haven't tried with larger systems yet.

Interestingly, seems like no extra compiler flags needed for OMP, though I'm not yet convinced it's actually working properly.... it seems to multi-thread only intermittently for me.


jflat06:
Closing this to clean things up.

PR #212, opened 2019-07-11
Aleaverfay/empty and zero

aleaverfay:
Allocate empty (uninitialized) arrays for the terms, and then launch a kernel to set the values of this array to zero. This is broadly faster than cudaMemset.


aleaverfay:
IIRC, further benchmarking revealed that this offered no improvement

PR #221, opened 2019-09-11
Aleaverfay/stacked performance

aleaverfay:
Beginning to tinker with performance following the addition of stacked scoring.

So far:
1. LK-isotropic dispatch over heavy atom pairs --> ~4x speedup for LK.
2. `score::common::accumulate<Real>::add_one_dst` performs a (safe) cooperative-group reduction before performing atomicAdd to get a 2x speedup.

I probably won't merge this branch and am only creating the PR so you two can look at what I've got so far.

### DIFF
--- a/tmol/kinematics/builder.py
+++ b/tmol/kinematics/builder.py
@@ -90,31 +90,54 @@ class KinematicBuilder:
         )
         return bonds_csr
 
+    @classmethod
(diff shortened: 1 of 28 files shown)


aleaverfay:
Long outdated code. Performance improvements were integrated into the res-centric scoring code.

PR #223, opened 2019-11-21
Aleaverfay/stack kinematics

aleaverfay:
Construct a kinematic tree for a stack of systems.

The kinematics code has a "flattened" view of the atoms in the system, with the system-indices of the kintree atoms being their row-major index;

ie. if there are 3 systems with at most 50 atoms each, then atom 5 in system 2 (indexing from zero) would be:

50 * 2 + 5

so that 

    coords = torch.zeros( (nstacks, system_size, 3), dtype=float)

can be written to by a view

    coords_flattened = coords.reshape((-1, 3))

Within the kintree, there's a single root at position 0 (as before) and this root connects to the first atom (atom 0) for each of the systems. Within the `KinematicBuilder` code, this is accomplished by creating faux bonds between the root of the first system to the roots of all the other systems, ahead of the call to `csgraph.breadth_first_order`. 

The messiest part of this conversion is that I'm tolerating the two descriptions of bonds:
as either a nbonds x 2 array of atom index pairs, or a nbonds x 3 array of (stack-index, at1, at2) triples. There are several access points, it seems, into the `KinematicBuilder` code, (at least as far as testing goes) and I've had to make Unions of the n x 2 and n x 3 inputs and then paste on a column of 0's in case the user has passed an n x 2 array (corresponding to "everything is in a single stack").

### DIFF
--- a/tmol/kinematics/builder.py
+++ b/tmol/kinematics/builder.py
@@ -90,31 +90,54 @@ class KinematicBuilder:
         )
         return bonds_csr
 
+    @classmethod
+    @validate_args
+    def faux_bonds_between_roots(
+        cls,
+        roots: NDArray(int)[:],
+        natoms_total: int
+    ) -> Union[sparse.spmatrix, int]:
+        """Construct a csgraph with edges from the first root to all
+        other roots, if there are other roots. Otherwise, return 0.
+        The size of this csgraph should be the total number of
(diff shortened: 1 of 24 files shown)


aleaverfay:
Very old code here; eventually bits and pieces of this branch were integrated into the res-centric code.

PR #224, opened 2020-07-15
Update numpy to 1.15+

aleaverfay:
This version offers access to a function, take_along_axis, which
is needed in branch aleaverfay/build_rotamers to sort the indices
of the rotamers in the input rotamer library by decreasing
probability.

Frustratingly, there is a bug in numpy above 1.14 that causes
calls to numpy.all and numpy.any which pass a torch tensor to
fail with the error:

```
TypeError: any() missing 1 required positional arguments: "dim"
```

as noted here: https://github.com/pytorch/pytorch/issues/15810

The workaround is to simply ask for the torch tensor as a numpy
array.

### DIFF
--- a/environments/core.dependencies.txt
+++ b/environments/core.dependencies.txt
@@ -1,6 +1,6 @@
 python>=3.6
 ipython
-numpy==1.14.*
+numpy==1.15.*
 pandas==0.22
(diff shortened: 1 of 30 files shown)


aleaverfay:
Long outdated.

PR #243, opened 2022-07-27
First pass of residue-centric architecture

aleaverfay:
The residue-centric architecture resolves parameters for the molecular system on a per-residue level in its inner-most subroutines. Whereas pre-scoring "rendering" of per-atom data (creating what Frank calls "denormalized" parameter arrays) alleviates an extra hop through memory during score function evaluation -- perhaps useful in repeated score evaluations of a single structure -- it means that code that attempts to score a structure or, e.g., rotamers that live outside of a structure, must be very different from the scoring code for repeated evaluation. In particular, there is no way to describe inter-residue bond distances for rotamer-pair-energy evaluations without appealing to the idea of a residue and the number of chemical bonds separating pairs of residues. There will not / cannot be a way to compute all-against-all chemical bond separations in the context of disembodied rotamers.

In the residue-centric version, each block of atoms is identified as belonging to a particular residue type (aka block type), and then properties of that block type are looked up. Parameters for the residue can be loaded in to shared memory, e.g., for residue i and also for residue j. Then threads can each take one atom-pair to focus on, load the parameters for that atom pair into local memory from shared memory, and compute the atom pair energy. Some of this data loading was entirely possible in the whole-system (denormalized) rendering scheme, but notably, chemical-bond-separation data cannot be loaded in a coherent way across multiple residues simultaneously, requiring that a scheme that first denormalizes residue-specific data into "soup of atoms" data re-normalize that data and treat residues as coherent units again.

This code also includes a fledgling FoldForest, a multiple-pose analog of a FoldTree, representing a "user friendly" interface for KinForest (formerly KinTree) construction.

This commit merges in a series of branches long-since diverged from trunk.


environments/core.dependencies.txt
```diff
@@ -0,0 +1,11 @@
+python>=3.6
```
fdimaio, line 1: latest merge should have removed these .txt files (only the yml files are present now)

tmol/database/chemical/__init__.py
```diff
@@ -56,36 +62,80 @@ class Connection:
...
+    mainchain_atoms: Optional[Tuple[str, ...]]
+    sidechain_chirality: str
+    termini_variants: Tuple[str, ...]
+
+
+@attr.s(auto_attribs=True, frozen=True, slots=True)
+class ProtonationProperties:
+    protonated_atoms: Tuple[str, ...]
+    protonation_state: str
+    pH: float
```
fdimaio, line 106: I assume this is to handle alternate protonations?   How does this work?

tmol/kinematics/builder.py
```diff
@@ -103,85 +217,378 @@ def bonds_to_connected_component(
...
+
+        n_kf_atoms = len(kfo_2_to)
+        n_roots = len(to_roots)
+        n_target_atoms = numpy.max(kfo_2_to) + 1
+
+        to_2_kfo = invert_mapping(kfo_2_to, n_target_atoms)
+        # to_2_kfo = numpy.full((kfo_2_to.max() + 1), -1, dtype=numpy.int32)
+        # to_2_kfo[kfo_2_to] = numpy.arange(n_kf_atoms, dtype=numpy.int32)
+
+        # Screw pandas
```
fdimaio, line 276: can you delete these lines

tmol/kinematics/builder.py
```diff
@@ -103,85 +217,378 @@ def bonds_to_connected_component(
...
+        kfo_parents[kfo_roots] = component_parent
+        kfo_roots += kin_start
+        frame_x += kin_start
+        frame_y += kin_start
+        frame_z += kin_start
+
+        def _t(x):
+            return torch.tensor(x, dtype=torch.int32)
+
+        # numpy.set_printoptions(threshold=10000)
```
fdimaio, line 348: delete these

tmol/kinematics/builder.py
```diff
@@ -103,85 +217,378 @@ def bonds_to_connected_component(
...
+            parent=_t(kfo_parents),
+            frame_x=_t(frame_x),
+            frame_y=_t(frame_y),
+            frame_z=_t(frame_z),
+        )
+
+        return attr.evolve(self, kinforest=cat((self.kinforest, extended_kin_forest)))
+
+    # unneeded @classmethod
+    # unneeded def count_uniq_atom_inds(cls, all_atoms: NDArray[numpy.int32][:]):
```
fdimaio, line 379: delete these

tmol/kinematics/compiled/compiled.cpu.cpp
```diff
@@ -98,40 +101,40 @@ struct InverseKinDispatch {
     auto dofs = dofs_t.view;
 
     auto k_coords2hts = ([=] EIGEN_DEVICE_FUNC(int i) {
-		if (i==0) {
```
fdimaio, line 101: i think this (and the failed linting) is due to a change in the version of clang-format. the current env.yml should have the right version.

tmol/kinematics/compiled/compiled.ops.cpp
```diff
@@ -0,0 +1,174 @@
...
+  return coords;
+
+};
+
+
+
+static auto registry =
+    torch::jit::RegisterOperators()
+        .op("tmol::forward_kin_op", &kinematic_op )
+        .op("tmol::forward_only_kin_op", &forward_only_op );
```
fdimaio, line 170: so rather than making these separate functions exposed to python, I think the proper "torch" thing to do is to check (in the torch C++ layer) if dof gradients are required, then dispatch to the appropriate C++ function.

tmol/kinematics/compiled/kernel_segscan.cuh
```diff
@@ -106,9 +107,9 @@ struct cta_lbs_segscan_t {
       // p[vt-1] checks if the current node (the last of the cta)
       //   is a segment node.  If so, no carryout.
       if (p[vt - 1]) {
-        carry_out_values[cta] = (type == scan_type_inc)
-                                    ? output[cur_item - 1]
-                                    : op(result.scan, x[vt - 1]);
+        // Note the carry out doesn't depend on the scan type!
+        carry_out_values[cta] =
+            carry_in ? op(result.scan, x[vt - 1]) : x[vt - 1];
```
fdimaio, line 112: good catch

tmol/kinematics/compiled/kernel_segscan.cuh
```diff
@@ -282,6 +283,19 @@ void kernel_segscan(
       count, segments, num_segments, cta_dim.nv(), mp_data, context);
   nvtx_range_pop();
 
+  // std::cout << std::flush << "load balance partitions\n";
```
fdimaio, line 286: remove

tmol/kinematics/datatypes.py
```diff
@@ -83,11 +115,39 @@ def node(
 
     @classmethod
     def root_node(cls):
-        """The global/root kinematic node at KinTree[0]."""
+        """The global/root kinematic node at KinForest[0]."""
         return cls.node(
             id=-1, doftype=NodeType.root, parent=0, frame_x=0, frame_y=0, frame_z=0
         )
 
+    # @classmethod
```
fdimaio, line 123: remove

tmol/pack/rotamer/build_rotamers.py
```diff
@@ -0,0 +1,1043 @@
...
+    orig_atom_offset_for_poses_blocks = (
+        (
+            poses.block_coord_offset.flatten()[real_poses_blocks]
+            + per_pose_offset[real_poses_blocks]
+        )
+        .cpu()
+        .numpy()
+    )
+
+    # print("orig atom offset for poses blocks")
```
fdimaio, line 460: remove

tmol/pack/rotamer/build_rotamers.py
```diff
@@ -0,0 +1,1043 @@
...
+#
+#
+# NOTE: sidechain samplers should be listed in the residue-level packer task
+# NOTE: all samplers should be given the same starting data and then a mask
+#       to indicate which residues they should operate on
+
+
+@validate_args
+def rebuild_poses_if_necessary(
+    poses: PoseStack, task: PackerTask
```
fdimaio, line 88: is it necessary to rebuild the pose for this functionality?  Can PBT either be held separate from pose or updated in a way that does not change derived data w/i posestack?

tmol/pack/rotamer/build_rotamers.py
```diff
@@ -0,0 +1,1043 @@
...
+@validate_args
+def create_dof_inds_to_copy_from_orig_to_rotamers(
+    poses: PoseStack,
+    task: PackerTask,
+    samplers,  # : Tuple[ChiSampler, ...],
+    rt_for_rot: Tensor[torch.int64][:],
+    block_type_ind_for_rot: Tensor[torch.int64][:],
+    sampler_for_rotamer: Tensor[torch.int64][:],
+    n_dof_atoms_offset_for_rot: Tensor[torch.int32][:],
+) -> Tuple[Tensor[torch.int64][:], Tensor[torch.int64][:]]:
```
fdimaio, line 586: wonder if it makes sense to move all the indexing fns to an indexing class?

tmol/pack/rotamer/chi_sampling.hh
```diff
@@ -0,0 +1,12 @@
+#pragma once
+
+namespace tmol {
+namespace pack {
+namespace rotamer {
+
+void num_possible_rots(
```
fdimaio, line 7: ?

tmol/pack/rotamer/dunbrack/dispatch.impl.hh
```diff
@@ -0,0 +1,840 @@
...
+          tmol::numeric::bspline::ndspline<2, 3, D, Real, Int>::interpolate(
+              rotprob_slice, bbdihe);
+      rotamer_probability[possible_rotamer] =
+          score::common::get<0>(prob_and_derivs);
+    };
+    Dispatch<D>::forall(
+        n_possible_rotamers, calculate_possible_rotamer_probability);
+
+    //     auto rot_prob_temp_tp = TPack<Real, 1,
+    //     tmol::Device::CPU>::zeros({rotamer_probability.size(0)}); auto
```
fdimaio, line 462: remove

tmol/pack/rotamer/mainchain_fingerprint.py
```diff
@@ -0,0 +1,352 @@
...
+    max_fingerprint: Tensor[torch.int32][:]
+
+
+@validate_args
+def create_non_sidechain_fingerprint(
+    rt: RefinedResidueType,
+    parents: NDArray[numpy.int32][:],
+    sc_atoms: NDArray[numpy.int32][:],
+    chem_db: ChemicalDatabase,
+):
```
fdimaio, line 111: Let's discuss this in-person (specifically chirality stuff)

tmol/pack/sim_anneal/annealer.py
```diff
@@ -0,0 +1,199 @@
...
+            self.alternate_coords,
+            self.alternate_coord_offsets,
+            self.alternate_block_id,
+            self.random_rots,
+            self.block_type_n_atoms,
+            self.max_n_atoms,
+        )
+
+        # context_block_type = context_block_type.to(torch.int64)
+        # n_contexts = self.pose_id_for_context.shape[0]
```
fdimaio, line 96: remove

tmol/pack/sim_anneal/compiled/annealer.cpu.cpp
```diff
@@ -0,0 +1,596 @@
...
+	score_events
+      );
+
+  sim_annealer->set_metropolis_accept_reject_step(metropolis_step);
+}
+
+
+    
+//   torch::Tensor context_coords,
+//   torch::Tensor context_block_type,
```
fdimaio, line 304: remove

tmol/pose/pose_kinematics.py
```diff
@@ -0,0 +1,453 @@
...
+    kin_polymeric_bonds = get_polymeric_bonds_in_fold_forest(
+        pose_stack, kin_polymeric_connections
+    )
+
+    # TO DO: add bonds between jump atoms
+    # TO DO: determine which atoms on a block a jump should
+    # connect to. Logic in R3: take the central "mainchain" atom
+    # which is only ok for polymers, but perverse for anything else.
+    # What's the mainchain of a ligand?!
+    # jump_atom_pairs = get_jump_bonds_in_fold_forest(pose_stack, fold_forest)
```
fdimaio, line 437: maybe atom with minimal max #bonds to any other atom?

tmol/pose/pose_stack.py
```diff
@@ -0,0 +1,716 @@
...
+
+    residue_coords: NDArray[numpy.float32][:, :, :, 3]
+
+    # coordinates are held as [n-poses x max-n-atoms x 3]
+    # where the offset for each residue are held in the
+    # block_coord_offset tensor [n-poses x max-n-blocks]
+    coords: Tensor[torch.float32][:, :, 3]
+
+    block_coord_offset: Tensor[torch.int32][:, :]
+    block_coord_offset64: Tensor[torch.int64][:, :]
```
fdimaio, line 42: just curious, why the distinction?  is it because torch needs int64 for indexing?  Why not just use int64?

tmol/score/dunbrack/potentials/compiled.hh
```diff
@@ -0,0 +1,116 @@
+#pragma once
+
```
fdimaio, line 2: this is duplicated from dispatch.hh?

tmol/score/common/accumulate.hh
```diff
@@ -13,6 +14,46 @@ namespace common {
 
 #define def auto EIGEN_DEVICE_FUNC EIGEN_STRONG_INLINE
 
+// #ifdef __CUDA_ARCH__
+#ifdef __CUDACC__
+
+// Perform reduction within the active threads.
```
fdimaio, line 20: cool!

tmol/score/ljlk/LJLKEnergy.py
```diff
@@ -0,0 +1,349 @@
...
+            type_params=self.type_params,
+            global_params=self.global_params,
+            lj_lk_weights=lj_lk_weights,
+        )
+
+    def create_block_neighbor_lists(
+        self, systems: PoseStack, system_bounding_spheres: Tensor[float][:, :, 4]
+    ):
+        # we need to make lists of all block pairs within
+        # striking distances of each other that we will use to
```
fdimaio, line 119: fine for now but in the future should think about moving most of this logic to common functions

fdimaio (approved):

tmol/database/chemical/__init__.py
aleaverfay: I added this to future-proof the addition of alternate protonation states to the residue-type set so that the Packer (the PackerPalette) has a way to say "I don't want to include this protonated ASP residue type in this packing run." The PackerPalette has the code now to handle later additions of alternate protonation states. I don't, however, have any tests to make sure that logic is working correctly.

tmol/kinematics/builder.py
aleaverfay: Done

tmol/kinematics/builder.py
aleaverfay: Done.

tmol/kinematics/builder.py
aleaverfay: Done

tmol/kinematics/compiled/compiled.cpu.cpp
aleaverfay: Turns out this was because we've never been linting .cpp files!

tmol/kinematics/compiled/compiled.ops.cpp
aleaverfay: ooooh, hmm

tmol/kinematics/compiled/kernel_segscan.cuh
aleaverfay: Done

tmol/kinematics/datatypes.py
aleaverfay: Done

tmol/pack/rotamer/build_rotamers.py
aleaverfay: This is the consequence of two design decisions:

The first is a rule for the `PackedBlockTypes` class is that new ResidueTypes / BlockTypes cannot be added to it after its creation. This allows annotations to be made to the PackedBlockTypes by name. E.g. Lennard Jones, by way of its base-class, the `AtomTypeDependentTerm`, makes the `atom_type` annotation on the PackedBlockType class, and then, when setting up for scoring, only checks if the annotation exists, but does not do any consistency checks.

The second is that `PackedBlockTypes` should be a data member of the `PoseStack` (as opposed to, say, global data), and the residue-type indices of the `PoseStack` depend on what block types exist in the `PoseStack`'s PackedBlockTypes object. Thus if there are only Leu block types in the PBT object before packing, and the packer wants to introduce other block types, then it needs to rebuild the pose with this alternate PBT. This is easy, however: besides the `block_type_ind`/`block_type_ind64 `tensors, the other tensors need only be reallocated and data copied into them.

Now, there's no rule that says that we can't initialize a `PoseStack` with the One True PBT that contained all the BTs needed for a particular trajectory. In that case, the rebuilding would be unnecessary. However, if we want to allow users to expand the set of BTs they allow over the course of a trajectory, then we would need to ensure the packer has the machinery necessary to rebuild the `PoseStack`.

fdimaio: OK, that makes sense.

tmol/pack/rotamer/build_rotamers.py
aleaverfay: What would that look like?

tmol/pack/rotamer/chi_sampling.hh
aleaverfay: lol, false start here

tmol/pack/rotamer/dunbrack/dispatch.impl.hh
aleaverfay: Done

tmol/pack/rotamer/build_rotamers.py
fdimaio: (say) A class rotdof2kfdof that held offsets and other info as class members?

tmol/pack/sim_anneal/annealer.py
aleaverfay: done

tmol/pack/sim_anneal/compiled/annealer.cpu.cpp
aleaverfay: Done

tmol/pose/pose_kinematics.py
aleaverfay: I'm going to come back to this one; yes, we need the ability to create jumps, and that isn't a low priority, but it's definitely not the top of the list.

tmol/pose/pose_stack.py
aleaverfay: I don't want to use int64 in our kernels because it's a waste of bandwidth and local memory / shared memory when we have no situations in which we will ever hit our heads on the 2 billion upper limit on the number of things represented by a signed integer. However, yes, torch wants 64 bit integers for all of their operations, and we still write a decent number of things using torch. So the choice is then either 1) keep int32 as a data member and convert to int64 only as often as its needed in a `block_coord_offset.to(torch.int64)` call or 2) keep both int32 and int64 around. It felt like the number of `.to` calls I needed was so high that it was wasteful to launch a kernel just to make one of these conversions and then to use the data a single time before throwing it away (only to re-create it again soon thereafter).

tmol/score/dunbrack/potentials/compiled.hh
aleaverfay: Looks like. I axed it just now along w/ compiled.pybind.cpp

PR #247, opened 2022-12-08
Aleaverfay/res centric elec

aleaverfay:
Create residue centric, whole-pose term for the electrostatics potential.

Note: both ref2015 and beta_nov16 use the "branch atom is dipole representative" behavior in Rosetta3 and this branch dutifully replicates that decision, even if it seems that the results from energy function parameterization showed that "base atom is dipole representative" behavior to have an edge.


fdimaio (approved):
LGTM

PR #249, opened 2023-02-10
Aleaverfay/res centric lkball

aleaverfay:
Res-centric implementation of the lk_ball energy term. The term loads tiles of atoms from residue 1 and residue 2 into shared memory and iterates across polar/occluder pairs, one thread per polar/occluder pair. The thread then evaluates all water/occluder and then water/water interactions. The kernels use lots of registers and the low occupancy is surely the source of the comparatively poor performance of this term compared to LJ/elec/hbonds. Various attempts to speed up this code by 1) load balancing water/occluder and water/water interactions or 2) loading fewer coordinates into registers at a time, have not yielded any performance benefits. Accepting this implementation and moving forward.

### DIFF
--- a/tmol/score/lk_ball/potentials/lk_ball.hh
+++ b/tmol/score/lk_ball/potentials/lk_ball.hh
@@ -5,6 +5,7 @@
 #include <cmath>
 
 #include <tmol/score/common/geom.hh>
+#include <tmol/score/common/accumulate.hh>
 #include <tmol/score/ljlk/potentials/lk_isotropic.hh>
 
 #include "params.hh"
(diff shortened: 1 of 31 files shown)


tmol/score/lk_ball/potentials/lk_ball.hh
```diff
@@ -478,8 +492,870 @@ struct lk_ball_score {
...
+  accum_derivs4(
+      occluder_block_dat, occluder_start + occluder_atom_tile_ind, dV.dJ);
+  for (int i = 0; i < MAX_N_WATER; ++i) {
+    water_accum_derivs4(
+        polar_block_dat, polar_start + polar_atom_tile_ind, i, dV.dWI);
+    water_accum_derivs4(
+        occluder_block_dat, occluder_start + occluder_atom_tile_ind, i, dV.dWJ);
+  }
+}
+
```
fdimaio, line 1265: maybe give better names that *4 / *1 ?

fdimaio (commented):
Looks good.  I see a lot of the same patterns emerging from the other two terms ... maybe it would be good to add a comment with high-level overview somewhere in scoring machinery?

fdimaio (approved):

tmol/score/lk_ball/potentials/lk_ball.hh
aleaverfay: Hmm. Gonna back-burner this name change suggestion and wrap up the PR but I'll make sure I get it on the next revision.

PR #250, opened 2023-03-07
Add 'new' omega score term. 

jflat06:
Initial commit just to get things in. Will add full functionality as implemented.


jflat06:
Getting some non-deterministic behavior.

Sometimes the first value in the 10-pose energies will not match any of the rest of them, and will cause the test to fail. Other times, the GPU portion segfaults. I'm getting these are related.

jflat06:
Suspecting that the non-deterministic behavior is happening at the last residue, since I do no checks to see if there is anything on the other end of the connection.

tmol/score/omega/params.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+@attr.s(auto_attribs=True, slots=True, frozen=True)
+class OmegaGlobalParams(TensorGroup):
+    K: Tensor[torch.float32][...]
+
+
+@attr.s(frozen=True, slots=True, auto_attribs=True)
+class OmegaParamResolver(ValidateAttrs):
+    """Container for global/type/pair parameters, indexed by atom type name.
```
aleaverfay, line 26: FWIW, the "ParamResolver" idiom is mostly a holdover from the PackedResidueSystem way of approaching scoring. The ParamResolvers would take a set of residues and create denormalized tensors of parameters so that each atom had its own parameters and didn't need to be thought of as part of a particular residue. I don't believe you need a param resolver for omega; the tensor of the GlobalParams can be created simply by the OmegaEnergyTerm in its __init__ method.

tmol/score/omega/potentials/omega.hh
```diff
@@ -0,0 +1,23 @@
...
+#include <tmol/score/common/tuple.hh>
+
+namespace tmol {
+namespace score {
+namespace omega {
+namespace potentials {
+
+#define def auto EIGEN_DEVICE_FUNC EIGEN_STRONG_INLINE
+
+using namespace tmol::score::common;
```
aleaverfay, line 16: I would like to avoid "using namespace" statements in header files

tmol/score/omega/potentials/compiled.ops.cpp
```diff
@@ -70,6 +72,116 @@ class ScoreOp : public torch::autograd::Function<ScoreOp<DispatchMethod>> {
...
+        ingrad = ingrad.unsqueeze(-1);
+      }
+
+      result.emplace_back(saved_grad * ingrad);
+    }
+
+    int i = 0;
+    auto dscore_dcoords = result[i++];
+
+    return {
```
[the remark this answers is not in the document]
jflat06: Hey Andrew, this code is very much unfinished - I was mostly checking it in to have somewhere to discuss issues/questions. I haven't done anything for the backward step at all yet, this is just copy pasted code from LJLK (I think).

tmol/score/omega/potentials/omega_pose_score.hh
```diff
@@ -0,0 +1,53 @@
...
+      TView<Int, 2, D> pose_stack_block_type,
+      TView<Vec<Int, 2>, 3, D> pose_stack_inter_block_connections,
+      TView<Int, 3, D> block_type_omega_quad_uaids,
+      TView<Int, 3, D> block_type_atom_downstream_of_conn,
+
+      TView<OmegaGlobalParams<Real>, 1, D> global_params,
+
+      bool compute_derivs
+
+      ) -> std::tuple<TPack<Real, 1, D>, TPack<Vec<Real, 3>, 2, D>>;
```
aleaverfay, line 47: The convention for pose scores is that each term should return a 2D tensor of values:
1st dimension: the number of sub-terms defined (e.g. lk-ball defines 4, fa_elec defines 1)
2nd dimension: the number of poses in the stack.
That way the set of 2D tensors can be concatenated along their first dimensions and then the weighted scores  computed by a matrix-vector dot product (see tmol/score/score_function.py:186)

jflat06: Ok, I was wondering about that, as the shape of the baseline I copied from LJLK didn't match. Thanks!

tmol/score/omega/potentials/omega_pose_score.impl.hh
```diff
@@ -0,0 +1,206 @@
...
+        pose_stack_block_coord_offset[pose_index][block_index];
+    return block_coord_offset + uaid[0];  // The actual global atom index
+  } else {                                // We need to follow to another block
+    int connection_index = uaid[1];
+    int sep = uaid[2];
+
+    const Vec<Int, 2>& connection =
+        pose_stack_inter_block_connections[pose_index][block_index]
+                                          [connection_index];
+    int other_block_index = connection[0];
```
aleaverfay, line 70: Yes, this may be -1 if there is no residue on the other side of a connection. This is actually how we handle termini right now. This function looks a lot like a function I've sketched (but not tested) for Rama. 

https://github.com/uw-ipd/tmol/blob/aleaverfay/res_centric_rama/tmol/score/rama/potentials/rama_pose_score.impl.hh#L174 

It would probably be wise to refactor both of these functions into a common location. In any case, the function should return -1 for any uaid that failed to resolve and then the calling function would need to look for that condition.

tmol/score/omega/potentials/omega_pose_score.impl.hh
```diff
@@ -0,0 +1,206 @@
...
+    int other_block_type_index =
+        pose_stack_block_type[pose_index][other_block_index];
+
+    int idx = block_type_atom_downstream_of_conn[other_block_type_index]
+                                                [other_connection_index]
+                                                [sep];  // The offset within the
+                                                        // other block
+    int block_coord_offset =
+        pose_stack_block_coord_offset[pose_index][other_block_index];
+    return block_coord_offset + idx;
```
jflat06, line 81: @aleaverfay This is where I suspect things might be producing non-deterministic behavior since I do nothing to check if a connection is valid.

tmol/score/omega/potentials/omega_pose_score.impl.hh
jflat06: Ok, great!

Is there a reason to prefer lambdas over declaring a function in the namespace like this?

jflat06: Any suggestions on a proper home for this code?

tmol/score/omega/potentials/omega.hh
jflat06: Huh, I think I coped it from another header file that I was using as an example. I've moved it to compiled.ops.cpp where it is needed.

tmol/score/omega/potentials/omega_pose_score.impl.hh
jflat06: Handling the -1 case _appears_ to have resolved the non-deterministic behavior, at least from my ~15 manual runs.

tmol/score/omega/omega_energy_term.py
```diff
@@ -0,0 +1,88 @@
...
+from tmol.score.common.stack_condense import tile_subset_indices
+from tmol.score.omega.params import OmegaGlobalParams
+from tmol.score.omega.omega_whole_pose_module import OmegaWholePoseScoringModule
+
+from tmol.chemical.restypes import RefinedResidueType
+from tmol.pose.packed_block_types import PackedBlockTypes
+from tmol.pose.pose_stack import PoseStack
+
+
+class OmegaEnergyTerm(AtomTypeDependentTerm, BondDependentTerm):
```
jflat06, line 16: Is this actual the correct usage of AtomTypeDependentTerm and BondDependentTerm? When are these necessary/what do they provide?

tmol/score/omega/omega_energy_term.py
```diff
@@ -0,0 +1,88 @@
...
+        self.tile_size = OmegaEnergyTerm.tile_size
+
+    @classmethod
+    def score_types(cls):
+        import tmol.score.terms.omega_creator
+
+        return tmol.score.terms.omega_creator.OmegaTermCreator.score_types()
+
+    def n_bodies(self):
+        return 2
```
jflat06, line 34: Technically the calculation involves 2 residues, but it is O(n) on the number of blocks. What should this be?

tmol/score/omega/omega_energy_term.py
```diff
@@ -0,0 +1,88 @@
...
+
+    def setup_block_type(self, block_type: RefinedResidueType):
+        super(OmegaEnergyTerm, self).setup_block_type(block_type)
+
+        if hasattr(block_type, "omega_quad_uaids"):
+            return
+
+        omega_quad_uaids = torch.full((4, 3), -1, dtype=torch.int32)
```
jflat06, line 42: Should I be putting these in a device torch tensor here? I think it's probably unnecessary because the setup_packed_block_types func makes a copy into a separate tensor regardless, and only that tensor is passed down into the kernel.

What is the purpose of the distinction between setup_block_type and setup_packed_block_types, then? Why not just do everything in setup_packed_block_types?

tmol/score/omega/params.py
```diff
@@ -0,0 +1,36 @@
...
+                return t
+
+        global_params = OmegaGlobalParams(
+            **{
+                n: at_least_1d(torch.tensor(v, device=device))
+                for n, v in cattr.unstructure(omega_database.global_parameters).items()
+            }
+        )
+
+        return global_params
```
jflat06, line 36: This is basically just packing the OmegaDatabase into tensors so they're available on the GPU. Does this look right?

From here, the OmegaWholePoseScoringModule labels these as parameters with the line: 

`self.global_params = _p(torch.stack(_t([global_params.K]), dim=1))`

which concatenates them all into a single tensor? Is this more complicated than it needs to be?

tmol/score/omega/potentials/compiled.ops.cpp
```diff
@@ -70,6 +74,99 @@ class ScoreOp : public torch::autograd::Function<ScoreOp<DispatchMethod>> {
...
+                  TCAST(block_type_atom_downstream_of_conn),
+                  TCAST(global_params),
+                  coords.requires_grad());
+
+          score = std::get<0>(result).tensor;
+          dscore_dcoords = std::get<1>(result).tensor;
+        }));
+#ifdef __CUDA_ARCH__
+    cudaDeviceSynchronize();
+#endif
```
jflat06, line 116: Remove before commit - was useful for getting printf() output from GPU.

tmol/score/omega/potentials/params.hh
```diff
@@ -43,4 +58,21 @@ struct enable_tensor_view<score::omega::potentials::OmegaParameters<Real>> {
...
+
+  static const int nconsumed_dims = 1;
+  static const int consumed_dims(int i) {
+    return (i == 0) ? sizeof(score::omega::potentials::OmegaGlobalParams<Real>)
+                          / sizeof(Real)
+                    : 0;
+  }
+
+  typedef typename enable_tensor_view<Real>::PrimitiveType PrimitiveType;
+};
```
jflat06, line 76: I just copy/pasted this from LJLK, and don't really know what's going on here. Are there any specific considerations? Is this code that can be factored out somehow?

tmol/tests/score/omega/test_omega_energy_term.py
```diff
@@ -0,0 +1,67 @@
...
+    omega_energy.setup_packed_block_types(p1.packed_block_types)
+    omega_energy.setup_poses(p1)
+
+    omega_pose_scorer = omega_energy.render_whole_pose_scoring_module(p1)
+
+    def score(coords):
+        scores = omega_pose_scorer(coords)
+        return torch.sum(scores)
+
+    gradcheck(score, (p1.coords.requires_grad_(True),), eps=1e-3, atol=1e-2, rtol=5e-3)
```
jflat06, line 40: How are these parameters? Any tweaks I should make?

tmol/tests/score/omega/test_omega_energy_term.py
```diff
@@ -0,0 +1,65 @@
+import numpy
+import torch
+
+from tmol.score.omega.omega_energy_term import OmegaEnergyTerm
+from tmol.pose.packed_block_types import residue_types_from_residues, PackedBlockTypes
+from tmol.pose.pose_stack_builder import PoseStackBuilder
+
+from tmol.tests.autograd import gradcheck
+
+
```
jflat06, line 10: What is the standard set of tests that I should implement?

tmol/tests/score/omega/test_omega_energy_term.py
```diff
@@ -0,0 +1,65 @@
...
+def test_smoke(default_database, torch_device: torch.device):
+
+    omega_energy = OmegaEnergyTerm(param_db=default_database, device=torch_device)
+
+    assert omega_energy.device == torch_device
+    assert omega_energy.global_params.K.device == torch_device
```
jflat06, line 16: Anything else I should test in here?

tmol/score/omega/potentials/omega_pose_score.impl.hh
```diff
@@ -0,0 +1,176 @@
...
+  auto V_t = TPack<Real, 2, D>::zeros({1, n_poses});
+  auto dV_dx_t = TPack<Vec<Real, 3>, 3, D>::zeros({1, n_poses, max_n_atoms});
+
+  auto V = V_t.view;
+  auto dV_dx = dV_dx_t.view;
+
+  // Optimal launch box on v100 and a100 is nt=32, vt=1
+  LAUNCH_BOX_32;
+  // Define nt and reduce_t
+  CTA_REAL_REDUCE_T_TYPEDEF;
```
jflat06, line 121: What do these do? Are these appropriate for omega?

tmol/score/omega/potentials/omega_pose_score.impl.hh
```diff
@@ -0,0 +1,176 @@
...
+
+    accumulate<D, Real>::add(V[0][pose_index], common::get<0>(omega));
+    for (int j = 0; j < 4; ++j) {
+      accumulate<D, Vec<Real, 3>>::add(
+          dV_dx[0][pose_index][omega_indices[j]], common::get<1>(omega).row(j));
+    }
+  });
+
+  int total_blocks = pose_stack_block_coord_offset.size(1);
+  DeviceDispatch<D>::forall_stacks(n_poses, total_blocks, func);
```
jflat06, line 168: Is this the correct dispatch that I should be using for omega?

jflat06:
I have implemented the following tests:

test_smoke
test_whole_pose_scoring_module_gradcheck_whole_pose
test_whole_pose_scoring_module_10

Are there other standard tests that I should include as well? Anything specific to omega?

jflat06:
@aleaverfay @fdimaio I've cleaned up the code and made a few comments/questions, if you have some time to take a look.

aleaverfay (changes requested):
This is looking really good. After you've cleaned up the linting (let me know if any part of that process is unclear), the handful of small changes I suggested, and fixed the "remove before merging" cudaDeviceSynchronize call, I think you'll be ready.

aleaverfay:
@fdimaio The codecov report isn't showing up for this PR. Is that hook broken? Could you have a look?

tmol/score/omega/params.py
jflat06: Doesn't the value K need to be converted to a tensor somewhere?

tmol/score/omega/potentials/omega_pose_score.impl.hh
jflat06: Is the 32-threads-per-CTA going to be spmething that is energy term specific? Do we want this centrally defined so that we can update it more easily depending on the backend hardware?

tmol/score/omega/potentials/compiled.ops.cpp
```diff
@@ -80,6 +177,7 @@ Tensor score_op(Tensor coords, Tensor params) {
 #define TORCH_LIBRARY_(ns, m) TORCH_LIBRARY(ns, m)
 TORCH_LIBRARY_(TORCH_EXTENSION_NAME, m) {
   m.def("score_omega", &score_op<common::ForallDispatch>);
+  m.def("score_omega_new", &omega_pose_scores_op<DeviceOperations>);
```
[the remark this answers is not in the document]
jflat06: Good catch! This was intended to be temporary, but somehow slipped by when I self-reviewed. I'll try to add more explicit TODOs in the future.

fdimaio:
> @fdimaio The codecov report isn't showing up for this PR. Is that hook broken? Could you have a look?

Looking now

tmol/database/scoring/__init__.py
```diff
@@ -23,6 +25,7 @@ class ScoringDatabase:
     def from_file(cls, path=os.path.dirname(__file__)):  # noqa
         return cls(
             ljlk=LJLKDatabase.from_file(os.path.join(path, "ljlk.yaml")),
+            omega=OmegaDatabase.from_file(os.path.join(path, "omega.yaml")),
```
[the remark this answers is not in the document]
jflat06: I've rearranged all of the entries so that everything is now in alphabetical order.

tmol/score/omega/potentials/omega_pose_score.impl.hh
jflat06: I have moved this to tmol/score/common/uaid_util.hh

jflat06:
> This is looking really good. After you've cleaned up the linting (let me know if any part of that process is unclear), the handful of small changes I suggested, and fixed the "remove before merging" cudaDeviceSynchronize call, I think you'll be ready.

Hey Andrew, I am assuming you're talking about the buildkite/tmol/microscope-linting check that is failing? I am unfamiliar with buildkite. I have made an account - do you need to add me to an orangization?

tmol/score/omega/omega_energy_term.py
jflat06: Something that occurred to me: what do we want to do for types that don't have omega defined? Should we fill the omega_quad_uaids for that block type as all -1? That should cause the uaid resolution function to properly skip this.

jflat06: Another quick question - does it make sense to extract the omega values out here, when they are already members of RefinedResidueType? They exist in the torsion_to_uaids Mapping as a Tuple instead of a numpy type, and only exist conditionally. Not sure where to draw the line at redundancy.

tmol/score/omega/params.py
jflat06: I'm inclined to leave this, so that it can serve as a more general example of how to include parameters for anyone wishing to write a score term in the future. If you'd prefer it be simplified, I can change it.

jflat06:
@aleaverfay - It seems like the only remaining issue is the codecov check, which is failing after the last commit: 

https://app.codecov.io/gh/uw-ipd/tmol/pull/250/commits

This was an attempt to alphabetize the benchmark tests. Is this a concern? Should I revert it?

jflat06:
@aleaverfay @fdimaio Looks like it passed!

aleaverfay:
> @aleaverfay @fdimaio Looks like it passed!

Fantastic! It looks like you'll have to resolve a conflict with trunk, but once you've done that and the tests come back clean, go ahead and squash merge. When you do that, write a solid description of your changes and make sure that the PR number is included in your commit message so that github will link it back to this branch. You can delete the branch once you've merged (I have been not good about that, personally).

jflat06:
Looks like it is blocking the merge since it is still flagged as 'changes requested'. @aleaverfay if things look good, do you want to approve it?

jflat06:
> > Looks like it is blocking the merge since it is still flagged as 'changes requested'. @aleaverfay if things look good, do you want to approve it?
> 
> It looks like linting is still failing (code formatting issues). Possible that you are running the wrong version of black or that the pre-commit hooks are not updating the failing file?

That was a problem previously when I didn't have the conda environment activated by accident. I'll double check that it is correct tomorrow.

jflat06:
Ah, I see what happened. I resolved the merge conflicts directly on github, which obviously didn't run the pre-commit steps. I'll make a note not to do that on this repo.

fdimaio (approved):
LGTM

tmol/score/common/uaid_util.hh
```diff
@@ -0,0 +1,65 @@
...
+template <typename Real, typename Int, tmol::Device D>
+TMOL_DEVICE_FUNC int resolve_atom_from_uaid(
+    TensorAccessor<Int, 1, D> uaid,
+    int block_index,
+    int pose_index,
+
+    TView<Int, 2, D> pose_stack_block_coord_offset,
+    TView<Int, 2, D> pose_stack_block_type,
+    TView<Vec<Int, 2>, 3, D> pose_stack_inter_block_connections,
+    TView<Int, 3, D> block_type_omega_quad_uaids,
```
jflat06, line 29: @aleaverfay I just noticed while searching for omega that this rogue argument made its way into this function. What's the best way to remove this? Should I make a mini-PR? Or do you just want to update this when you use this function for your task that needed it?

PR #252, opened 2023-03-21
Aleaverfay/res centric rama

aleaverfay:
This implements the Ramachandran (rama_prepro) term in a residue-centric way. This is a two-body energy as which table from residue i depends on the amino acid identity of residue i+1. The scores match the baseline from R3.


tmol/score/rama/rama_energy_term.py
```diff
@@ -0,0 +1,140 @@
...
+                    )
+        rama_torsion_atoms = rama_torsion_atoms.reshape(8)
+
+        upper_conn_id = numpy.full((1,), -1, dtype=numpy.int32)
+        if "up" in block_type.connection_to_cidx:
+            upper_conn_id[0] = block_type.connection_to_cidx["up"]
+        is_pro = numpy.full((1,), 0, dtype=numpy.int32)
+
+        # TO DO: Better logic here to handle proline variants
+        if block_type.name == "PRO":
```
fdimaio, line 81: we need to think about this (and it arises much worse in fa_dun).  I think there needs to be two parts: 1) what does this score term consider a proline?, and 2) how are patched residue types specified?  The answer to 1 is probably different for different score terms.

This is fine for now though.

fdimaio (approved):
LGTM.  Very simple implementation.

PR #253, opened 2023-03-21
Add disulfide score term

jflat06:
WIP

Currently adds a new PDB to the database that contains several disulfides. 

Finds the disulfides according to proximity and adds the appropriate residue connections.

Flags connections as 'disulfide' with a new tensor in the packed block types. 

Extracts the (hopefully) appropriate atoms from the two disulfided bonded residues. 

Currently working on scoring the actual bond, but there are some missing math functions.


jflat06:
Currently a mess because it includes the omega stuff, which should be sorted out once it makes its way into master.

PR #254, opened 2023-03-22
Implement disulfide score term

jflat06:
Introduces the disulfide bond score term. 

The EnergyTerm itself down into the C++ code should be complete. 

Limitations:

The loading of the actual disulfides from the test fixtures, all the way up to the annotation of the blocks may require further work in the future, depending on how we want to handle this. 

Currently the way of loading and annotating the disulfides in the block types allows for only 1 disulfide per block type. If we develop a system for describing multiple disulfide conenctions per block type, only the annotation step will need to change. Everything else in the EnergyTerm should work fine with multiple disulfide bonds per block.


tmol/score/disulfide/disulfide_energy_term.py
```diff
@@ -0,0 +1,84 @@
...
+        import tmol.score.terms.disulfide_creator
+
+        return tmol.score.terms.disulfide_creator.DisulfideTermCreator.score_types()
+
+    def n_bodies(self):
+        return 1
+
+    def setup_block_type(self, block_type: RefinedResidueType):
+        super(DisulfideEnergyTerm, self).setup_block_type(block_type)
+
```
jflat06, line 38: Not currently doing anything here, should probably fill the block types with data on the CPU side here.

tmol/score/disulfide/disulfide_energy_term.py
```diff
@@ -0,0 +1,84 @@
...
+            (packed_block_types.n_types, packed_block_types.max_n_conn),
+            0,
+            dtype=torch.int32,
+            device=self.device,
+        )
+
+        for i, bt in enumerate(packed_block_types.active_block_types):
+            # TODO: hacking this in for now
+            print("RESIDUE: " + bt.name)
+            if bt.name == "CYD":
```
jflat06, line 61: Obviously a total hack that should be fixed.

tmol/score/disulfide/disulfide_energy_term.py
```diff
@@ -0,0 +1,86 @@
...
+        if hasattr(block_type, "disulfide_connections"):
+            return
+
+        disulfide_connections = numpy.array([], dtype=numpy.int32)
+        if "dslf" in block_type.connection_to_cidx.keys():
+            disulfide_connections = numpy.append(
+                disulfide_connections, [block_type.connection_to_cidx["dslf"]]
+            )
+
+        setattr(block_type, "disulfide_connections", disulfide_connections)
```
jflat06, line 49: Right now the block types are just filled with a numpy array with the connection indices of the any disulfide connections. 

@aleaverfay Does this sound like an appropriate way of annotating the disulfides?

The current structure of the RefinedResidueTypes only really allows for 1 disulfide connection currently, but if that changes, this is the place where we'd update the code to use that.

tmol/score/disulfide/disulfide_energy_term.py
```diff
@@ -0,0 +1,86 @@
...
+        disulfide_conns = torch.full(
+            (packed_block_types.n_types, packed_block_types.max_n_conn),
+            0,
+            dtype=torch.int32,
+            device=self.device,
+        )
+
+        for i, bt in enumerate(packed_block_types.active_block_types):
+            for conn in bt.disulfide_connections:
+                disulfide_conns[i, conn] = 1
```
jflat06, line 69: Here's where we make the actual tensor to pass into the CUDA code. It's just a n_block_type*n_max_conns array that flags each connection as a disulfide or not. I should probably update this to be a bool instead of int32.

jflat06:
Unlike omega, this is an entirely new score term. Is there any additional work that needs to be done to hook this into everything? I did a grep for omega and it seemed like all other code using omega was using the old omega.

tmol/database/default/scoring/disulfide.yaml
```diff
@@ -0,0 +1,28 @@
...
+  dss_mixed_mu2 : 90.0
+  dcs_logA1 : -15.8644
+  dcs_mu1 : -1.26015618528
+  dcs_kappa1 : 13.3778
+  dcs_logA2 : -16.9017
+  dcs_mu2 : 1.36188623499
+  dcs_kappa2 : 13.6370
+  dcs_logA3 : -7.0219
+  dcs_mu3 : -3.0115756416500004
+  dcs_kappa3 : 2.9327
```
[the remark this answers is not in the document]
jflat06: I've removed all the terms that were unused (they all had 'mixed' in the name). https://github.com/uw-ipd/tmol/pull/254/commits/68e31d89e1940888ca376d3c64c9fe020afd4257

tmol/score/disulfide/disulfide_energy_term.py
jflat06: Updated to use bool.

tmol/score/disulfide/potentials/potentials.hh
```diff
@@ -0,0 +1,209 @@
...
+
+  const Real MEST = exp(-20.0);
+  const Real WT_DIH_SS(0.1);
+  const Real WT_DIH_CS(0.1);
+  const Real WT_ANG(0.1);
+  const Real WT_LEN(0.1);
+  const Real SHIFT(2.0);
+
+  const Real res1_d_multiplier(1.0);
+  const Real res2_d_multiplier(1.0);
```
jflat06, line 64: Rosetta has considerations for different chiralities. I did not include that functionality, but I left these variables in here. Should I remove them, or should we leave them and add that code later?

tmol/score/disulfide/potentials/potentials.hh
```diff
@@ -0,0 +1,209 @@
...
+      dihedral_angle<Real>::V_dV(block1_CA, block1_CB, block1_S, block2_S);
+  auto disulf_ca_dihedral_angle_2 =
+      dihedral_angle<Real>::V_dV(block2_CA, block2_CB, block2_S, block1_S);
+
+  const Real MEST = exp(-20.0);
+  const Real WT_DIH_SS(0.1);
+  const Real WT_DIH_CS(0.1);
+  const Real WT_ANG(0.1);
+  const Real WT_LEN(0.1);
+  const Real SHIFT(2.0);
```
jflat06, line 61: Not sure which of these should be included in the disulfide DB. They were private class variables in Rosetta, and I just copied their default values.

tmol/tests/data/__init__.py
```diff
@@ -62,6 +62,13 @@ def ubq_system():
     return read_pdb(pdb.data["1ubq"])
 
 
+@pytest.fixture(scope="session")
+def disulfide_res():
```
jflat06, line 66: Named this 'disulfide' since that probably is a more useful name to anyone in the future looking for a fixture with disulfides in it.

tmol/database/default/scoring/disulfide.yaml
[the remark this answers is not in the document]
fdimaio: mixed ones are for mixed D/L ... we will need them once D AAs are in

jflat06:
@fdimaio Shortening the disulfide params lead to a change of:

```
E       AssertionError: 
E       Not equal to tolerance rtol=1e-05, atol=1e-05
E       
E       Mismatched elements: 10 / 10 (100%)
E       Max absolute difference: 0.00015092
E       Max relative difference: 4.633452e-05
E        x: array([[-3.257311, -3.257311, -3.257311, -3.257311, -3.257311, -3.257311,
E               -3.257311, -3.257311, -3.257311, -3.257311]], dtype=float32)
E        y: array([[-3.25716, -3.25716, -3.25716, -3.25716, -3.25716, -3.25716,
E               -3.25716, -3.25716, -3.25716, -3.25716]], dtype=float32)
```

So pretty minimal. Does this look OK?

tmol/score/disulfide/potentials/params.hh
```diff
@@ -0,0 +1,120 @@
...
+
+  Real wt_dih_ss;
+  Real wt_dih_cs;
+  Real wt_ang;
+  Real wt_len;
+  Real shift;
+};
+
+template <typename Real, tmol::Device D>
+struct DisulfideGlobalParamTensors {
```
aleaverfay, line 51: It doesn't look like this class is necessary. I don't see it used anywhere else.

aleaverfay (changes requested):
This is looking really good! I have a few very minor changes.

tmol/chemical/restypes.py
```diff
@@ -368,3 +368,26 @@ def find_simple_polymeric_connections(
...
+def find_disulfide_connections(res: List[Residue],) -> List[Tuple[int, str, int, str]]:
+    residue_connections = []
+
+    cystines = [
+        (ind, cys) for ind, cys in enumerate(res) if cys.residue_type.name == "CYD"
+    ]
+    for i, cys1 in cystines:
+        for j, cys2 in cystines:
+            if i < j:
+                ca_index = cys1.residue_type.atom_to_idx["SG"]
```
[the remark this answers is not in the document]
jflat06: Doh!

tmol/score/disulfide/disulfide_energy_term.py
```diff
@@ -0,0 +1,83 @@
...
+                disulfide_connections, [block_type.connection_to_cidx["dslf"]]
+            )
+
+        setattr(block_type, "disulfide_connections", disulfide_connections)
+
+    def setup_packed_block_types(self, packed_block_types: PackedBlockTypes):
+        super(DisulfideEnergyTerm, self).setup_packed_block_types(packed_block_types)
+
+        if hasattr(packed_block_types, "disulfide_conns"):
+            return
```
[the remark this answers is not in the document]
jflat06: Something like this? https://github.com/uw-ipd/tmol/pull/254/commits/99023c48306a08c616d2f38c0071aae5c5ff2d37

It feels a little weird to be putting this in the packed block types, since it isn't a feature of the block types and this data won't change if the active block types change. We could consider putting it in DisulfideEnergyTerm::\_\_init\_\_ and have it done once right after it loads the params from the database.

tmol/score/disulfide/disulfide_whole_pose_module.py
```diff
@@ -0,0 +1,75 @@
...
+        def _t(ts):
+            return tuple(map(lambda t: t.to(torch.float), ts))
+
+        self.pose_stack_block_coord_offset = _p(pose_stack_block_coord_offset)
+        self.pose_stack_block_types = _p(pose_stack_block_types)
+        self.pose_stack_inter_block_connections = _p(pose_stack_inter_block_connections)
+        self.bt_disulfide_conns = _p(bt_disulfide_conns)
+        self.bt_atom_downstream_of_conn = _p(bt_atom_downstream_of_conn)
+
+        self.global_params = _p(
```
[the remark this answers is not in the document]
jflat06: Something like this? https://github.com/uw-ipd/tmol/pull/254/commits/99023c48306a08c616d2f38c0071aae5c5ff2d37

It feels a little weird to be putting this in the packed block types, since it isn't a feature of the block types and this data won't change if the active block types change. We could consider putting it in DisulfideEnergyTerm::\_\_init\_\_ and have it done once right after it loads the params from the database.

tmol/score/disulfide/disulfide_energy_term.py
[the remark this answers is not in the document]
jflat06: If I understood you correctly, something like this? https://github.com/uw-ipd/tmol/pull/254/commits/17390b62a134432186c2f4dfcfcbc04c1caa9204

Should I be doing this test for all cached data members?

aleaverfay (approved):
lgtm

jflat06:
Great, I'll wait for the checks to pass and then merge if they check out.

jflat06:
> I just changed the interface to the resolve_uaid function in the rama PR (#252) to take UAIDs as a struct instead of as a TensorAccessor. That doesn't affect this PR as far as I can tell. The dislufide_pose_score.impl.hh file does include uaid_util.hh, but doesn't invoke the functions it contains. It most certainly will affect the cart-bonded term, though, so I wanted to give you the head's up, @jflat06

I don't think this actually affects the cart-bonded term. Cart-bonded forgoes the UAID system because it needs access to non-mainline atoms coming from a connection. So part of the implementation was adding a 'paths starting from connection' data structure to each block type.

PR #256, opened 2023-04-24
Aleaverfay/py3.9 torch1.11

aleaverfay:
Update the build environment for tmol to python3.9 and some of the more recent versions of many packages. This will help us use tmol along side other recent packages.


aleaverfay:
@fdimaio Could you look at the racehorse-benchmark? I am not sure I understand why it's showing as failing.

fdimaio:
> @fdimaio Could you look at the racehorse-benchmark? I am not sure I understand why it's showing as failing.

The script makes some plots at the end and uses 'seaborn' as a dependency to do so.  It's missing in the current environment.

I personally do not find the plots useful, so you can either remove the plotting or add seaborn to deps.

fdimaio (approved):
Mostly looks fine.  I approved, but before merging can you do the following:
1) Can you fix the seaborn issue (see comment)?
2) Can you remove the 'test' subfolder?

fdimaio:
looks like it needs itermplot too

PR #257, opened 2023-05-09
Aleaverfay/pose building interface

aleaverfay:
This PR creates an interface to tmol so that NN packages can output coordinates and residue-type information for run-of-the-mill proteins in a canonical form and then construct a PoseStack object needed for score function evaluation. There are a few features/drawbacks here worth mentioning:

1) chemical type resolution for HIS and CYS is handled by this code
2) if hydrogen atoms are not provided, their positions can be computed
3) termini variants are not yet handled
4) the HIS and CYS chemical resolution steps are handled on the CPU and will later become faster

Details:
HIS chemical type resolution:
Histadine (at neutral pH) exists in one of two tautomerization states (90% of the time): 1) where the NE2 nitrogen is protonated and the ND1 nitrogen is not, and 2) where the ND1 nitrogen is protonated and the NE2 nitrogen is not. There are four possible ways that tmol will resolve this ambiguity:
   1) the user provides either the HD1 or the HE2 hydrogen coordinates (but not both) and tmol selects the corresponding tautomerization state
   2) the user provides a coordinate for "HN," a stand in for the hydrogen that is chemically bound to one of the HIS's two nitrogens; tmol will use the distance of HN to ND1 and to NE2 to and will declare the tautomerization state based on the shorter of the two
   3) the user provides the coordinates for three atoms "HN", "NH" and "NN" which are stand-ins for the hydrogen attached to the nitrogen, the nitrogen attached to the hydrogen, and the nitrogen that is not attached to the hydrogen. The distances between NH and NN and the ring's CG atom are used to decide which of the two tautomerization states the ring is in.
   4) the user provides only the heavy-atom coordinates, in which case the code chooses the NE2-is-protonated tautomerization state.

CYS disulfide detection:
Disulfides are detected by distance between SG atoms with a maximum detection distance of 2.5A. The algorithm is greedy: cysteine residues are iterated across in order in an outer loop (i) and again in an inner loop (j). The closest distance between SG on CYS i and any other as-of-yet-unpaired CYS j within 2.5A is found and a disulfide bond declared between i and j. This does not guarantee that the largest number of possible disulfides is found; if res 1 and 3 have a distance of 2.2A and 1 and 2 has 2.4A, and 3 and 4 have a distance of 2.4A, e.g. then the largest number of disulfides would be formed when 1 and 2 are paired and 3 and 4 are paired. This algorithm, however, will pair residues 1 and 3. Also of note: if the order of the residues changes, but their distances do not, so that 1 and 2 have a distance of 2.4A and then 2 and 3 have the distance of 2.2A, (and cysteines 3 and 4 continue to have a distance of 2.4A) and for sake of argument the other pairs have distances > 2.5, then the 1-2 disulfide will be found first, and therefore the 3-4 disulfide will also be found later.

It is currently not possible to tell tmol which pairs of CYS residues should be disulfide bonded, but will be in the future.

H-placement:
If hydrogen atoms are missing, they will be built from ideal coordinates wrt heavy atoms. This is a fully-differentiable process so derivatives wrt the score function that apply to the hydrogen atoms get converted into derivatives for the atoms that define their coordinates. In the case of HIS, the NE2-protonated tautomerization state is chosen; in the case of SER, THR, CYS, and TYR, the hydroxyl's dihedral angle is assigned to 180. Future versions of this code will make more intelligent decisions on how to choose the dihedral angle, e.g. by looking for nearby hydrogen bond acceptors.

Currently there is no way to build missing heavy atom coordinates, even if their position could be inferred from other heavy atom coordinates, e.g. for backbone oxygen atoms.


tmol/io/details/select_from_canonical.py
```diff
@@ -0,0 +1,430 @@
...
+    #   - res_not_connected[p, i, 1] must be false
+    #   - res_not_connected[p, i+1, 0] must be false
+    # - If a residue is the first residue in a chain and res_not_connected[p, i, 0]
+    #   is true, then it will not be treated as an n-terminal residue, rather,
+    #   it will be treated as a "mid" residue (unless it is also a c-term residue)
+    #   and, in the case of amino acids, will have a regular H atom bonded to N
+    #   instead of the 1H, 2H, and 3H atoms. Its connection to the "previous" residue
+    #   will be incomplete (-1) instead of to a particular residue.
+    # - If a residue is the last residue in a chain and res_not_connected[p, i, 1]
+    #   is true, then it will not be treated as a c-terminal residue.
```
fdimaio, line 84: Andrew, I don't like the "overloading" of res_not_connected and the interplay between chain and res_not_connected here.

Can we have an input Lx2 tensor to directly generate terminal variants? (e.g. if x[i,0]==true then i is an nterm variant)

tmol/io/details/select_from_canonical.py
```diff
@@ -0,0 +1,430 @@
...
+    return (block_type_ind64, inter_residue_connections64, ibb64)
+
+
+@validate_args
+def take_block_type_atoms_from_canonical(
+    packed_block_types: PackedBlockTypes,
+    block_types64: Tensor[torch.int64][:, :],
+    coords: Tensor[torch.float32][:, :, :, 3],
+    atom_is_present: Tensor[torch.int32][:, :, :],
+):
```
fdimaio, line 265: why is "canonical" here?  is this code reliant only on canonical AAs?

tmol/io/pose_stack_construction.py
```diff
@@ -0,0 +1,195 @@
...
+from tmol.pose.pose_stack import PoseStack
+
+
+def pose_stack_from_canonical_form(
+    chain_id: Tensor[torch.int32][:, :],
+    res_types: Tensor[torch.int32][:, :],
+    coords: Tensor[torch.float32][:, :, :, 3],
+    atom_is_present: Optional[Tensor[torch.int32][:, :, :]] = None,
+    disulfides: Optional[Tensor[torch.int64][:, 3]] = None,
+    res_not_connected: Optional[Tensor[torch.bool][:, :, 2]] = None,
```
fdimaio, line 13: i think a terminal-variant type might make more sense here? (see above)

tmol/io/pose_stack_construction.py
```diff
@@ -0,0 +1,195 @@
...
+    if atom_is_present is None:
+        atom_is_present = torch.all(torch.logical_not(torch.isnan(coords)), dim=3)
+
+    # 1
+    # this will return the same object each time to minimize the number
+    # of calls to the setup_packed_block_types annotation functions
+    pbt, atom_type_resolver = default_canonical_packed_block_types(chain_id.device)
+
+    # 2
+    # "left justify" the input canonical-form residues: residues that are
```
fdimaio, line 121: finish comment?

tmol/io/canonical_ordering.py
```diff
@@ -0,0 +1,977 @@
...
+
+
+max_n_canonical_atoms = max(
+    len(atlist) for _, atlist in ordered_canonical_aa_atoms.items()
+)
+
+
+def canonical_form_from_pdb_lines(pdb_lines, device):
+    ##### TEMP!!!! #####
+    # USE PDB V2 FOR NOW!
```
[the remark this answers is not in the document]
fdimaio: Can you document the output of this function?

tmol/io/canonical_ordering.py
```diff
@@ -0,0 +1,977 @@
...
+        "HA",
+        "1HB",
+        "2HB",
+        "HD1",
+        "HD2",
+        "HE1",
+        "HE2",
+        "HH",
+    ),
+}
```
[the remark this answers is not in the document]
fdimaio: Alternately can this be built/derived from the existing database?

tmol/io/canonical_ordering.py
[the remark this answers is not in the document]
fdimaio: also perhaps "canonic ordering" should be an input to this function?

tmol/io/chain_deduction.py
```diff
@@ -0,0 +1,139 @@
...
+from tmol.pose.pose_stack import PoseStack
+from tmol.pose.packed_block_types import PackedBlockTypes
+from tmol.types.functional import validate_args
+from tmol.types.array import NDArray
+from tmol.utility.ndarray.common_operations import exclusive_cumsum1d
+
+
+@validate_args
+def chain_inds_for_pose_stack(
+    pose_stack: PoseStack,
```
fdimaio, line 13: This is inferring chain id (for writing) from connected components in a pose stack?  You had mentioned this might be made user overridable at some point?

tmol/tests/conftest.py
```diff
@@ -26,6 +26,8 @@
     water_box_system,
     water_box_res,
     systems_bysize,
+    pertuzumab_lines,
+    pert_and_nearby_erbb2,
```
fdimaio, line 30: rather than give ids of the proteins, maybe give something more inline with the aspect they are testing? (antibody, antibody_and_nearby_antigen). 'pert_and_nearby_erbb2' does not mean much to me

tmol/score/score_function.py
```diff
@@ -187,4 +187,4 @@ def __init__(
 
     def __call__(self, coords):
         all_scores = torch.cat([term(coords) for term in self.term_modules], dim=0)
-        return torch.sum(self.weights * all_scores)
+        return torch.sum(self.weights * all_scores, dim=0)
```
fdimaio, line 190: what is the reason for this change?

tmol/io/pose_stack_construction.py
```diff
@@ -0,0 +1,195 @@
...
+        coords,
+        atom_is_present,
+        disulfides,
+        res_not_connected,
+    ) = left_justify_canonical_form(
+        chain_id, res_types, coords, atom_is_present, disulfides, res_not_connected
+    )
+
+    # 3
+    found_disulfides, res_type_variants = find_disulfides(res_types, coords, disulfides)
```
fdimaio, line 134: ok, if i've read the rest of this correctly, this function takes input disulfides and then runs detection looking for additional ones.  There should be a way to specify that detection shouldn't add new dslfs as well (e.g., even though SGs are close, NO dslfs should be "found")

fdimaio (changes requested):
Looks pretty good.  I added a few comments, general issues are:
- a few suggestions for generalizing the code so that it is easier to reuse the code when adding additional residues to "canonic set"
- the use of numba over pure torch for dslf/his detection?  could be parallelized very simply and would save lots of lines of code
- a few areas code commenting/asserts might help

tmol/io/details/build_missing_leaf_atoms.py
```diff
@@ -0,0 +1,314 @@
...
+                phi_backup += j_icoor.phi
+        elif not icoor_at_is_h(j_icoor.great_grand_parent):
+            # Logic for handling when the heavy-atom great-grand parent,
+            # which itself is a leaf atom, is absent. This "general" logic
+            # is specifically for building the OXT atom on a cterm residue
+            # when the O atom is given but OXT is not.
+            while icoor_at_is_leaf(j_icoor.great_grand_parent):
+                ggp_ind_backup = bt.icoors_index[j_icoor.great_grand_parent]
+                j_icoor = bt.icoors[ggp_ind_backup]
+                phi_backup += j_icoor.phi
```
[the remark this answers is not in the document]
fdimaio: can you add asserts or documentation of:
- what missing atoms it handles
- what happens if other atoms are missing

tmol/__init__.py
```diff
@@ -1,3 +1,14 @@
...
+from tmol.io.canonical_ordering import (  # noqa: F401
+    ordered_canonical_aa_atoms,
+    ordered_canonical_aa_atoms_v2,
+    max_n_canonical_atoms,
+    canonical_form_from_pdb_lines,
+)
+from tmol.io.pose_stack_construction import pose_stack_from_canonical_form  # noqa: F401
+from tmol.score import beta2016_score_function  # noqa: F401
+
+
```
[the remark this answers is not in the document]
aleaverfay: I like that idea. Let's start thinking about exactly what we want in our API besides what I have here and make that its own PR. I feel like there's a good bit of brainstorming that we should do and could handle independently of this PR

tmol/io/canonical_ordering.py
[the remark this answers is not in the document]
aleaverfay: > also perhaps "canonic ordering" should be an input to this function?

as in a boolean selecting between v2 and v3? or one of the "canonical-ordering" dictionaries themselves?

tmol/io/pose_stack_construction.py
aleaverfay: 🤦‍♂️

tmol/io/chain_deduction.py
aleaverfay: Yes, the main interface function for writing a PDB file for a PoseStack now allows the user to specify their own chain indices

tmol/tests/conftest.py
aleaverfay: I think as a name for a fixture, it makes sense to say "what is this?" and not "what is it good for in terms of testing." The unit tests that rely on these fixtures, though, should say "here is what I will be testing", e.g. "dslf-detection" or "chopped-up-chain-representation"

tmol/__init__.py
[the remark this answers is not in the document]
fdimaio: agreed

tmol/io/canonical_ordering.py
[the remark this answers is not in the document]
fdimaio: I meant the dictionary.

tmol/score/score_function.py
fdimaio: oh i see, got it

tmol/tests/conftest.py
aleaverfay: I renamed these "pertuzumab_pdb" and "pertuzumab_and_nearby_erbb2_pdb_and_segments"

environments/devel/Dockerfile
```diff
@@ -4,7 +4,12 @@ FROM nvidia/cuda:11.8.0-devel-ubuntu20.04 as base
 # general environment for docker
 ENV DEBIAN_FRONTEND=noninteractive
 
-RUN apt-get update; apt-get install -y --no-install-recommends sudo wget curl git vim rsync && rm -rf /var/lib/apt/lists/*
+RUN apt-get update; apt-get install -y --no-install-recommends sudo wget curl git vim rsync libxml2 && rm -rf /var/lib/apt/lists/*
+
+
+# Genentech users: uncomment the line below to download Roche certificates so that you can download from nvidia
+# RUN wget -r -l1 --no-parent --quiet -A .crt -P /usr/local/share/ca-certificates https://certinfo.roche.com/rootcerts/ && update-ca-certificates
+
```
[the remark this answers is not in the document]
aleaverfay: :+1:

tmol/io/canonical_ordering.py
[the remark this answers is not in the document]
aleaverfay: I want to punt the project of creating a new CanonicalForm / better PDB parsing down the road. I think this idea is important, but also want to see this current task wrapped up.

tmol/io/details/build_missing_leaf_atoms.py
```diff
@@ -0,0 +1,314 @@
...
+@validate_args
+def _annotate_packed_block_types_atom_is_leaf_atom(
+    pbt: PackedBlockTypes, atom_type_resolver: AtomTypeParamResolver
+):
+    if hasattr(pbt, "is_leaf_atom"):
+        # TO DO: it feels like "is_hydrogen" is the kind of data member
+        # that PBT ought to provide itself and does not belong as an
+        # annotation that the missing-leaf-atom-construction code
+        # ought to create.
+        assert hasattr(pbt, "is_hydrogen")
```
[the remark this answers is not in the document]
aleaverfay: ok, I noticed a design flaw as I tried to remove this logic; basically, a RawResidueType/RefinedResidueType has its atom types listed as strings and you cannot do anything with those strings in the absence of a ChemicalDatabase. They are essentially pointers and they cannot be dereferenced on their own. To make sense of this data, then, the objects that contain ResidueTypes need to hold a pointer to the ChemicalDatabase that defines their properties. So I've added those pointers to ResidueTypeSet and to PackedBlockTypes.

FWIW: These two classes do not need to hold only ResidueTypes that are defined within the ChemicalDatabase; they can hold ResidueTypes that are created programmatically, for instance. However, it does mean that all ResidueTypes within these two classes should be defined wrt a single ChemicalDatabase. That feels entirely reasonable to me.

aleaverfay:
FWIW, the "x" that's on codecov is from my disabling the unit testing of the packer / simulated annealing code. That code is not exactly dead, but it needs a lot of work before it's useful. The other code in the PR is pretty well covered.

fdimaio (approved):
Looks fine.  Need to update the atom ordering code but that is for a future update

PR #259, opened 2023-06-13
Fix pip installs

fdimaio:
Pip installs now copy all necessary files.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,3 +1,0 @@
-include LICENSE
-include README.md
-recursive-include tmol *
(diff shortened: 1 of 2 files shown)


kleinhenz:
I think using `find_packages` is slightly more idiomatic than putting everything in the manifest.

```
from setuptools import setup, find_packages

setup(
    name='myPackage',
    packages=find_packages(),
)
```

fdimaio:
> 



> I think using `find_packages` is slightly more idiomatic than putting everything in the manifest.

Ok, will update this.

fdimaio:
OK, fixed.

aleaverfay (approved):
I can confirm that this fixes the `pip install .` installation :+1:

PR #260, opened 2023-06-16
Residue centric cartbonded

jflat06:
This PR reworks cartbonded to use the new res-centric paradigm.

The overall flow of the calculation is as follows:

* Generate a list of angles/lengths/torsions (subgraphs) for each block type.
* Read the parameters from the cartbonded database and associate them with the block types. 
* Pack all of these together into tensors. The subgraphs are simply concatenated while the parameters are stuffed into a hash table with the keys being tuples of unique atom ids, and the values being the parameter values. 
* Pass this off to the C++.
* For each residue, iterate over all subgraphs (angles/lengths/torsions) for that block type, get the unique atom ids of the atoms involved, and do a lookup in the hash table to see if there is a set of parameters for that group of atoms. If there is, accumulate it onto the result.
* Also iterate over the inter-residue subgraphs and accumulate the results of any matching set of atoms. 

To accomplish the above, several systems were added:

* atom_unique_ids were added to the AtomTypeDependentTerm. AtomTypeDependentTerm iterates over all block types and generates a table with entries for the res-type specific atom id (ALA:C) and the wildcard variant (WILDCARD:C). This way every atom has a unique ID in a global indexing scheme across all atoms in the pose. These unique ids are used to form the keys in our hash table, which lets us do lookups in the C++ code based on the types of the atoms involved in a particular subgraph.
* atom_paths_from_conn were added to RefinedResidueType. This is similar to the existing atom_downstream_of_conn, except that it provides the full graph coming from that connection rather than just the mainline atoms. This is needed because some of our cartbonded terms apply to residue-spanning torsions that involve non-mainline atoms. 

Lastly, I changed the database format. I have changed the old database to have an "old" suffix. It should be trivial to remove when we remove non-res-centric code.


tmol/score/cartbonded/potentials/potentials.hh
```diff
@@ -38,6 +38,56 @@ def cblength_V_dV(Real3 atm1, Real3 atm2, Real K, Real x0)
   return {E, dE * dist.dV_dA, dE * dist.dV_dB};
 }
 
+template <typename Real>
+def cblength_V_dV2(Real3 atm1, Real3 atm2, Real K, Real x0)
+    ->tuple<Real, Vec<Real3, 2>> {
+  auto dist = distance<Real>::V_dV(atm1, atm2);
+  Real E = 0.5 * K * square(dist.V - x0);
+  Real dE = K * (dist.V - x0);
+  return {E, {dE * dist.dV_dA, dE * dist.dV_dB}};
```
jflat06, line 47: This is what is causing CUDA grief. This compiles and runs just fine on CPU. CUDA complains:

```
E   /home/jflat06/rosetta/tmol/tmol/score/cartbonded/potentials/potentials.hh(49): error: no instance of constructor "mgpu::tuple<args_t...>::tuple [with args_t=<float, Eigen::Matrix<Eigen::Matrix<float, 3, 1, 0, 3, 1>, 2, 1, 0, 2, 1>>]" matches the argument list
E               argument types are: (float, {...})
E             detected during:
E               instantiation of "auto tmol::score::cartbonded::potentials::cblength_V_dV2(tmol::score::cartbonded::potentials::Vec<Real, 3>, tmol::score::cartbonded::potentials::Vec<Real, 3>, Real, Real)->mgpu::tuple<Real, tmol::score::cartbonded::potentials::Vec<tmol::score::cartbonded::potentials::Vec<Real, 3>, 2>> [with Real=float]" 
```

Does CUDA not like the nested Eigen::Matrices? Or does it just fail to handle the nested {} notation?

fdimaio: Pretty sure it's the nested {}.  You need to initialize return val as Vec<Real3, 2> and use .col() to set each column separately.  e.g., in kinematic code:
>     ht.col(3) = Eigen::Matrix<Real, 1, 4>(0, 0, 0, 1);

tmol/score/cartbonded/cartbonded_energy_term.py
```diff
@@ -0,0 +1,285 @@
...
+        # Calculate the total number of params
+        total_params = sum(
+            [len(bt.cartbonded_params) for bt in packed_block_types.active_block_types]
+        )
+        # get the params not associated with any specific residue
+        wildcard_params = self.get_params_for_res("wildcard").items()
+        total_params += len(wildcard_params)
+
+        # stupid hash function
+        def hash_fun(key, max_size):
```
fdimaio, line 222: one potential source of error is that this needs to be synched w/ c++ version.  Maybe add a comment here about that (already 1 in cpp code).  Is it possible to have the magic numbers (0x1234, 3141) come from a single source?

tmol/tests/score/cartbonded/test_cartbonded_energy_term.py
```diff
@@ -0,0 +1,122 @@
...
+        cartbonded_energy.setup_block_type(bt)
+    cartbonded_energy.setup_packed_block_types(pbt)
+    """
+    assert hasattr(pbt, "cartbonded_quad_uaids")
+    assert pbt.cartbonded_quad_uaids.device == torch_device"""
+
+
+def test_whole_pose_scoring_module_gradcheck(
+    rts_ubq_res, default_database, torch_device: torch.device
+):
```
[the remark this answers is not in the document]
jflat06: Is it not scoped based on file? The other energy_term tests seem to use the generic function name, so I was trying to follow the convention (I did this for the other energy terms I have implemented as well).

tmol/score/cartbonded/cartbonded_energy_term.py
jflat06: The hash function code itself also must be synced, so I was thinking it was better to abstract it at the functional level. I'm not sure what the best way to share the magic numbers would be. You could always do a text file, or pass these values in. Those seem kind of hacky, though.

tmol/score/cartbonded/potentials/potentials.hh
jflat06: Yeah that was what I was avoiding, as I think it makes things less readable. I'll give it a try.

jflat06:
A design question for either of you - 

Right now I kept the atom identifiers in the yaml file the same - they just use the generic atom names. This means that I have to do the work to make the unique (res-specific) atom IDs and the wildcard IDs. It also means that I have to deduce when a param crosses the residue boundary. (C-N)

One could imagine an alternative where we have a way of specifying the unique atom IDs directly (or wildcard ID, if you wanted it to remain a wildcard). This would be a bit more explicit, and you could essentially load the params directly from the DB into the hash table. I'm not sure whether this is actually better or worth the effort to re-write it, though.

tmol/score/cartbonded/cartbonded_energy_term.py
jflat06: I've added a comment to the hash function here to address this.

tmol/score/cartbonded/cartbonded_energy_term.py
```diff
@@ -0,0 +1,287 @@
...
+            comb = list(permutations(bondmap[atom3], 3))
+            for atom1, atom2, atom4 in comb:
+                improper.append((atom1, atom2, atom3, atom4))
+
+        return (
+            lengths,
+            angles,
+            torsions,
+            improper,
+        )
```
jflat06, line 91: @aleaverfay 

This code is triggering the flake8 C901 for being too complex.

I'm not sure splitting this out into multiple functions would make it any more clear. Can I add an exception for this?

jflat06:
@aleaverfay this is passing all tests now, if you want to take a look.

aleaverfay (changes requested):
This looks really really good, Jeff. I've asked for just a handful of minor changes.

tmol/score/cartbonded/cartbonded_energy_term.py
```diff
@@ -0,0 +1,287 @@
...
+                hash_keys[index][i] = k
+            hash_keys[index][4] = max_value_index
+
+            for i, value in enumerate(values):
+                hash_values[max_value_index][i] = value
+
+        # Construct the params hash with the given scaling factor
+        SCALE = 2
+        hash_keys = numpy.full(
+            (total_params * SCALE, 5),
```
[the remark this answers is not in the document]
jflat06: Java (and I think several other languages) use a load factor of 0.75 for rehashing the table, so 0.5 should be a reasonable value.

I ran some stats on the hash table:

size: 2260
collisions: 609
average seek: 0.538938

The last value is probably the most meaningful. Basically, on average, we 'linear probe' an extra 0.53 times per hash lookup.

Bumping the scale factor up to 4x reduces this value even further to ~0.15. 

Really just comes down to time vs space, but the extra lookup time should be pretty negligible anyways (it basically is just the cost of comparing the key again, you don't need recompute the hash at all).

tmol/score/cartbonded/potentials/cartbonded_pose_score.impl.hh
```diff
@@ -0,0 +1,367 @@
...
+}
+
+// From a Vec of atom indices and a lookup table, return a new Vec with their
+// IDs preserving (-1)s
+template <typename Int, Int size, tmol::Device D>
+TMOL_DEVICE_FUNC Vec<Int, size> get_atom_ids(
+    TensorAccessor<Int, 1, D> atom_id_table, Vec<Int, size> atoms) {
+  Vec<Int, size> atom_ids;
+  for (int i = 0; i < size; i++) {
+    atom_ids[i] = get_atom_id<Int, D>(atom_id_table, atoms[i]);
```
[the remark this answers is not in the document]
jflat06: I've had some issues with the compiler complaining about not being able to deduce types, so I was trying to be explicit while debugging. 

I'll go ahead and remove these.

tmol/score/cartbonded/potentials/cartbonded_pose_score.impl.hh
```diff
@@ -0,0 +1,367 @@
...
+          }
+        });
+
+    if (subgraph_index >= subgraph_offset_next) {
+      if (subgraph_index >= subgraph_offset_next + NUM_INTER_RES_PATHS) return;
+
+      subgraph_index -= subgraph_offset_next;
+
+      // Iterate over each connection in this block
+      for (int i = 0; i < pose_stack_inter_block_connections.size(2); i++) {
```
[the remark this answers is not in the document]
jflat06: I'll go ahead and add a comment, but basically it is because atom order matters (Cα–C'–N–Cα is different from Cα–N–C'–Cα), and we do not check the reverse path for inter-residue paths, only for intra.

tmol/chemical/restypes.py
```diff
@@ -178,6 +178,54 @@ def _setup_path_distance(self):
...
+            paths[0] = (atom0, -1, -1)
+
+            idx = 1
+            for atom1 in bondmap[atom0] + [-1] * (3 - len(bondmap[atom0])):
+                if atom1 != -1:
+                    # atoms.append((atom0, atom1, -1))
+                    paths[idx] = (atom0, atom1, -1)
+                idx += 1
+
+            for atom1 in bondmap[atom0] + [-1] * (3 - len(bondmap[atom0])):
```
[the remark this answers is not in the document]
jflat06: I'm not really sure I understand.

There is the potential for a full row of -1s between valid paths, but I think this is necessary if we want to avoid having to iterate through tensor to find the locations of specific paths down in the C++. 

As you say, the point of the system is to keep the 1-atom paths, 2-atom paths, 3-atom paths in deterministic locations. 

If there is a simpler way of doing this, I can make modifications.

tmol/score/cartbonded/cartbonded_energy_term.py
```diff
@@ -0,0 +1,285 @@
...
+                        continue
+                    for atom4 in bondmap[atom3]:
+                        if atom2 == atom4:
+                            continue
+                        if atom1 >= atom4:
+                            continue
+                        torsions.append((atom1, atom2, atom3, atom4))
+
+        # get improper torsions
+        for atom3 in [block_type.atom_to_idx["CA"]]:
```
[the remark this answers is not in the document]
jflat06: Unless we want to include all possible improper torsions, I think we must be explicit. Are you suggesting that we should include all possible improper torsions? Or is there some way of determining which ones might be relevant that I'm not seeing?

That's a good question about residue types with no CA. I'm pretty sure this is just a python dict underneath, so it would give a key error. I can fix that.

tmol/score/cartbonded/cartbonded_energy_term.py
jflat06: That's a great point, I did not think about putting python code in that directory. I see now that there are some examples of that.

tmol/chemical/constants.py
```diff
@@ -16,3 +16,10 @@
 # computing this distance exactly is slow) without affecting the score
 
 MAX_SIG_BOND_SEPARATION = 6
+
+# MAX_PATHS_FROM_CONNECTION:
+#
+# The maximum number of paths coming from a particular block type's connection.
+# Currently this is 13 - 1 for the connection atom itself, 3 for the up-to-3 neighbors
```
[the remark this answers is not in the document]
jflat06: Yeah, that was a poor use of a dash.

I've updated with your suggestion and also added your generalized solution to the comment in case we want to utilize that in the future.

tmol/tests/score/cartbonded/test_cartbonded_energy_term.py
```diff
@@ -0,0 +1,132 @@
...
+
+    coords = torch.nn.Parameter(pn.coords.clone())
+    scores = cartbonded_pose_scorer(coords)
+
+    # make sure we're still good
+    torch.arange(100, device=torch_device)
+
+    numpy.testing.assert_allclose(
+        gold_vals, scores.cpu().detach().numpy(), atol=1e-5, rtol=1e-5
+    )
```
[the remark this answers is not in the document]
jflat06: How does this look? https://github.com/uw-ipd/tmol/pull/260/commits/74367b4589a48bfa7bfe61a164d765e3cd255d17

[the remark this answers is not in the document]
aleaverfay: 👍

aleaverfay (approved):
lgtm

tmol/chemical/restypes.py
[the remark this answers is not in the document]
jflat06: I've added a comment to try and make what is happening here more clear.

aleaverfay:
I think after you merge trunk into this branch and resolve the conflicts, that this PR is ready to be merged. 💪

tmol/score/atom_type_dependent_term.py
```diff
@@ -29,13 +33,54 @@ def __init__(self, param_db: ParameterDatabase, device: torch.device):
...
+        bt_atom_unique_id_index = numpy.full((len(atom_names)), -1, dtype=numpy.int32)
+
+        for index, unique_id_index in enumerate(unique_id_indices):
+            bt_atom_unique_id_index[index] = unique_id_index
+
+        for index, name_index in enumerate(name_indices):
+            bt_atom_name_index[index] = name_index
+
+        setattr(block_type, "atom_unique_ids", bt_atom_unique_id_index)
+        setattr(block_type, "atom_wildcard_ids", bt_atom_name_index)
```
jflat06, line 72: How will this react to MET and MET:nterm being in the RTS together?

jflat06:
> I think after you merge trunk into this branch and resolve the conflicts, that this PR is ready to be merged. 💪

Yeah, I am having some issues after resolving the conflicts due to the patching system changes. 

The first issue is that it was using the fullname to do lookups into the cartbonded DB, which was causing a keyerror since MET:nterm doesn't exist. I'm assuming we want to lookup from the basename on this? It should also probably not choke if that key doesn't exist.

The second issue is that the score is now wrong. 

```
E        x: array([[ 37.7848],
E              [183.5777],
E              [ 50.5842],...
E        y: array([[310.2358  ],
E              [194.5125  ],
E              [ 50.5842  ],...
```

I'm not 100% sure what's going on with this yet, but I think it could be due to the atom ID tables doubling up on the terminal entries. https://github.com/uw-ipd/tmol/pull/260/files#r1322083779

tmol/score/atom_type_dependent_term.py
jflat06: Seems like possibly it is sufficient to simply check if the atom_unique_id is in the index yet and don't add it if it is. This gets the scores very close to their previous values (though still failing).

jflat06: The above actually doesn't seem sufficient because the atoms in one variant may not match the ones in another. How do we deal with this? We can't tag them with the full name because then they would have separate IDs and the params wouldn't match (unless you added entries for the nterm variants in the cartbonded DB, which seems very silly). 

Seems like maybe you want to annotate the unpatched variant first, and then for any atoms in other variants that don't exist in the original, tag them as new atoms? I'm not really sure here.

aleaverfay:
Yes, it took me a little while to hunt it down, but the logic in Rosetta3 is to use the base_name to do the parameter lookups. I think the only atoms you should insert into the table are for the un-patched residue types; i.e. name == base_name

jflat06:
> Yes, it took me a little while to hunt it down, but the logic in Rosetta3 is to use the base_name to do the parameter lookups. I think the only atoms you should insert into the table are for the un-patched residue types; i.e. name == base_name

Ok, yeah. Frank and I talked about this and came up with a similar conclusion. However, my new baselines don't match the ones that Frank computed with his patching-aware cartbonded using the old implementation, so we weren't 100% sure which one was correct. Frank seemed to think it was his code that was at fault, though.

jflat06:
> Yes, it took me a little while to hunt it down, but the logic in Rosetta3 is to use the base_name to do the parameter lookups. I think the only atoms you should insert into the table are for the un-patched residue types; i.e. name == base_name

@aleaverfay 

Quick question - 

Right now I am adding ALL atoms under the base type name as they come in. So even if you see a variant type first, it will add the atoms then, including any variant-added atoms. Any atoms removed in the variant would be added when you ran the base type.

This has the side effect that some of the variant-only atoms get tagged under the base name. This could cause collisions if you had multiple variant types that add an atom with the same name. But that shouldn't matter if you aren't adding params that reference these atoms.

This is what I have implemented currently. 

If we want to ONLY tag the atoms in the base type, it gets a bit trickier.

One option is to somehow figure out which atoms are in the base type as you parse the variant, and add them under the basename.

If you don't do this, you need to ensure that:

1. The base type is in the RTS.
2. The base type is annotated first (before the variant). 

I don't know if those are safe assumptions. 

Any thoughts on this?

tmol/score/atom_type_dependent_term.py
```diff
@@ -29,13 +33,57 @@ def __init__(self, param_db: ParameterDatabase, device: torch.device):
...
+    def get_atom_wildcard_id_name(self, block_name, atom_name):
+        return "WILDCARD_ID:" + atom_name
+
+    def add_atom_index(self, block_type):
+        atom_names = [x.name for x in block_type.atoms]
+
+        unique_id_indices = []
+        name_indices = []
+        for atom_name in atom_names:
+            unique_id = self.get_atom_unique_id_name(block_type.base_name, atom_name)
```
aleaverfay, line 53: I've been puzzling over your "quick" question from last week surrounding the issue of name collision. Here, regardless of the order in which the block types are encountered, all the atoms that will ever be encountered will be given a unique atom id. Atoms that are part of n-term patches will get one id; atoms that are part of the c-term patches will get other ids; atoms that are part of both will get the same (individually unique) ids. The order in which atoms are processed will determine which atom IDs they are given, but the actual values of the IDs themselves make no difference (besides, perhaps some small effect on performance due to differences in hash function collisions). 

In the cart-bonded code where there are two different atoms with the same name that are created by different patches of the same base type, we would have "name collision" issues, yes. I think in that case, we should instead detect when two sets of parameters are specified for the same set of atoms and provide a warning. If (:pray: @everyday847) there are two patches that both add some "CM" atom -- one on a mono-methylated n-term nitrogen, and another one a di-methylated n-term nitrogen, and let's say that the patches both try to add a (different)"wildcard-N + wildcard+CM" set of bond-length parameters, then the cart-bonded database ought to stop execution and insist that either the atom names be changed in the patches to avoid the collision, or one of the patches ought to be converted into a new set of base types.

Therefore, I think that this strategy for lookup based on base-type name is the right way to go.

aleaverfay (approved):
Re: the possibility of name collision

Let's either add a TO DO note in the code that reads in cart-bonded parameters from the .yaml file to look for duplicated sets of atoms (i.e. duplicated keys) and to emit an error if the parameters (i.e. values) are not identical so that we can return to that task later, or, go ahead and add that check / error message. That way, if we leave this name-collision bear trap in the woods that tmol's users will at least be alerted when they've stepped in it.

tmol/score/cartbonded/potentials/cartbonded_pose_score.impl.hh
```diff
@@ -0,0 +1,392 @@
...
+TMOL_DEVICE_FUNC void accumulate_result(
+    common::tuple<Real, Vec<Real3, N>> to_add,
+    Vec<Int, N> atoms,
+    Real& V,
+    TensorAccessor<Vec<Real, 3>, 1, D> dV) {
+  accumulate<D, Real>::add(V, common::get<0>(to_add));
+  for (int i = 0; i < N; i++) {
+    accumulate<D, Vec<Real, 3>>::add(dV[atoms[i]], common::get<1>(to_add)[i]);
+  }
+}
```
jflat06, line 123: Here's an example of calling non-lambdas. Still suffers from not being able to capture parent scope, but maybe that would be less of an issue with packing args/vars into structs?

PR #262, opened 2023-06-29
Fix bug with strides of 0 for non-Scalar enable_tensor_views of tensors

aleaverfay:
In its backward-pass, torch.sum(t) returns a scalar, t2, which masquerades as a tensor with the same shape as t, except its strides will give it away as not being a real tensor: they will all be 0. But, if t is a tensor of coordinates and we use TCAST to coerce the t2 tensor into a TView<Eigen::Matrix<float, 3, 1>, N, D>, then the fact that dE_dx and dE_dy and dE_dz all have the same address (because a stride of 0 means "ignore the index in the given dimension" and a tensor with only strides of zeros means "ignore all the indices used to look up data, there is only one value") will be LOST and the dE_dy and dE_dz values will be read off the end of the array! Oh no!

To fix this, we can simply look at any consumed dimension as determined by the enable_tensor_view<SomeClass<T>> class and assert that the consumed dimension(s) has (have) a non-zero stride. Any other stride besides zero is fine.

### DIFF
--- a/tmol/tests/kinematics/test_script_modules.py
+++ b/tmol/tests/kinematics/test_script_modules.py
@@ -49,7 +49,7 @@ def test_kinematic_torch_op_backward_benchmark(benchmark, ubq_system, torch_devi
     kop = KinematicModule(tkinforest, torch_device)
 
     refold_kincoords = kop(tdofs.raw)
-    total = refold_kincoords.sum()
+    total = torch.sum(refold_kincoords[:, :])
 
     @benchmark
     def refold_grad():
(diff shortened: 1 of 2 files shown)


fdimaio (approved):
lgtm

aleaverfay:
It occurs to me that the non-zero stride is a necessary but not sufficient condition for well-allocated tensor casts; if the strides do not line up with the size of the object that we're casting to, e.g. if we have a stride of N != 1 for the last dimension of a coordinate tensor, e.g. an [n-pose x max-n-atoms x 3 x 20] coords tensor that we sliced as coords[:, :, :, 7] for the eighth set of coordinates, then when we cast to an Eigen::Matrix<Real, 3, 1>, we will actually read the X coordinates for copy 7, 8, and 9, instead of the X, Y and Z coordinates of the 7th copy. I think we should actually be asserting that the strides are the cumprod(j=0, j< i - 1, size[j]), i.e. that there is a contiguous block of memory that we are casting to

PR #263, opened 2023-07-10
Fix out-of-bounds bug in omega that produces NaNs in jagged PosesStacks

aleaverfay:
The score terms need checks for block types with sentinel values of `-1` so that jagged PoseStacks do not produce out-of-bounds reads / writes. This PR adds a unit test for omega scoring with jagged PoseStacks


fdimaio (approved):
lgtm

PR #265, opened 2023-08-01
Fix score function factory behavior for hbonds

aleaverfay:
Just an absolutely egregious copy-and-paste error: the logic for activating a score did not previously check to make sure that the EnergyTerm that gets activated is the one actually responsible for calculating energies for the score-type whose weight is being set. So I am adding that assertion statement as well.


fdimaio (approved):
lgtm

PR #266, opened 2023-08-03
Fix an out-of-bounds read in LKBall

aleaverfay:
I mistook the dimension ordering in the backward pass for lk-ball. I discovered this by adding boundary assertions to TensorAccessor.h and then doing a lot of printf debugging.

We really need a way to conditionally compile in boundary-check statements into one of the unit test executions, e.g. the cpu-only version.


fdimaio (approved):
Looks good to me.  A bounds check should not be hard to add for unit tests only.

PR #267, opened 2023-08-14
Fix bug in unresolved atom id resolution

aleaverfay:
The unresolved-atom-id resolution code should also handle cases when UAIDs specify the "this is not a real UAID" condition. Previously, it just accessed out of bounds memory locations. Oops.

I have now added unit tests for UAID resolution as well. This test hits the case where the old resolve_atom_from_uaid function would have read out of bounds (though, that would not necessarily have triggered a failure).

### DIFF
--- a/tmol/tests/score/common/test_uaid_util.py
+++ b/tmol/tests/score/common/test_uaid_util.py
@@ -1,17 +1,86 @@
 import numpy
 import torch
+import pytest
 
 from tmol.tests.score.common.uaid_util import resolve_uaids
 
 
-def test_resolve_uaids_smoke():
+@pytest.fixture
+def uaid_pose_stack():
+    n_blocks = 6
+
+    pose_stack_block_type = torch.zeros((1, n_blocks), dtype=torch.int32)
+
+    pose_stack_block_coord_offset = torch.zeros((1, n_blocks), dtype=torch.int32)
+    pose_stack_block_coord_offset[0, 0] = 0
+    pose_stack_block_coord_offset[0, 1] = 20
+    pose_stack_block_coord_offset[0, 2] = 40
+    pose_stack_block_coord_offset[0, 3] = 60
+    pose_stack_block_coord_offset[0, 4] = 80
+    pose_stack_block_coord_offset[0, 5] = 100
+
+    pose_stack_inter_block_connections = torch.zeros(
+        (1, n_blocks, 2, 2), dtype=torch.int32
+    )
+    # first residue
+    pose_stack_inter_block_connections[0, 0, 0, 0] = -1
+    pose_stack_inter_block_connections[0, 0, 0, 1] = -1
+    pose_stack_inter_block_connections[0, 0, 1, 0] = 1
+    pose_stack_inter_block_connections[0, 0, 1, 1] = 0
+    # 2nd residue
+    pose_stack_inter_block_connections[0, 1, 0, 0] = 0
+    pose_stack_inter_block_connections[0, 1, 0, 1] = 1
+    pose_stack_inter_block_connections[0, 1, 1, 0] = 2
+    pose_stack_inter_block_connections[0, 1, 1, 1] = 0
+    # 3rd residue
+    pose_stack_inter_block_connections[0, 2, 0, 0] = 1
+    pose_stack_inter_block_connections[0, 2, 0, 1] = 1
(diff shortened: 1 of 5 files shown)


jflat06:
Fix looks good!

For the unit test, I can't help but feel that this is treating a symptom rather than the cause.

In my mind, this bug made it in because the new code being added (omega) was not being tested on variety of inputs.

Could we perhaps enhance the 'test_whole_pose_scoring_module_single' (or _10, or add a new '_diverse') functions to test on a variety of structures rather than just UBQ1? For example, even just adding a disulfide structure would expose issues with variable numbers of connections. 

This would increase the burden of defining baselines, but it seems worth it to me. I guess we could have a "score_smoke" test that doesn't compare the baselines and is just there to catch proper errors, at least.

tmol/tests/score/common/test_uaid_util.py
```diff
@@ -0,0 +1,398 @@
...
+    # 5th residue
+    pose_stack_inter_block_connections[0, 4, 0, 0] = 3
+    pose_stack_inter_block_connections[0, 4, 0, 1] = 1
+    pose_stack_inter_block_connections[0, 4, 1, 0] = 5
+    pose_stack_inter_block_connections[0, 4, 1, 1] = 0
+    # 6th residue
+    pose_stack_inter_block_connections[0, 5, 0, 0] = 4
+    pose_stack_inter_block_connections[0, 5, 0, 1] = 1
+    pose_stack_inter_block_connections[0, 5, 1, 0] = -1
+    pose_stack_inter_block_connections[0, 5, 1, 1] = -1
```
jflat06, line 377: Seems like a lot of these definitions could be shared between the different tests.

tmol/tests/score/common/uaid_util.pybind.cc
```diff
@@ -0,0 +1,72 @@
...
+        pose_stack_block_coord_offset,
+        pose_stack_block_type,
+        pose_stack_inter_block_connections,
+        block_type_atom_downstream_of_conn);
+  }
+
+  return outputs_t.tensor;
+}
+
+PYBIND11_MODULE(TORCH_EXTENSION_NAME, m) {
```
jflat06, line 54: Do we want to be using different binding interfaces, or should we focus on one for uniformity?

jflat06 (approved):
All minor things, looks good to me.

tmol/tests/score/common/test_uaid_util.py
```diff
@@ -0,0 +1,398 @@
...
+    n_uaids = 4
+    uaids = torch.full((n_uaids, 3), -1, dtype=torch.int32)
+    block_inds = torch.zeros((n_uaids,), dtype=torch.int32)
+    pose_inds = torch.zeros((n_uaids,), dtype=torch.int32)
+    pose_stack_block_coord_offset = torch.zeros((1, n_blocks), dtype=torch.int32)
+    pose_stack_block_type = torch.zeros((1, n_blocks), dtype=torch.int32)
+    pose_stack_inter_block_connections = torch.zeros(
+        (1, n_blocks, 2, 2), dtype=torch.int32
+    )
+    block_type_atom_downstream_of_conn = torch.zeros((1, 2, 20), dtype=torch.int32)
```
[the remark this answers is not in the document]
aleaverfay: In this test, yes; that's because uaids[:,2] is 0

tmol/tests/score/common/test_uaid_util.py
```diff
@@ -0,0 +1,398 @@
...
+
+    uaids[0, 1] = 1
+    uaids[1, 1] = 1
+    uaids[2, 1] = 1
+    uaids[3, 1] = 1
+    uaids[0, 2] = 0
+    uaids[1, 2] = 0
+    uaids[2, 2] = 0
+    uaids[3, 2] = 0
```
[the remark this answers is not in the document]
aleaverfay: :+1:

tmol/tests/score/common/test_uaid_util.py
aleaverfay: yes, created a fixture

tmol/tests/score/common/uaid_util.pybind.cc
aleaverfay: This is sufficient for testing

PR #268, opened 2023-08-16
Aleaverfay/fix dun score termini

aleaverfay:
There were two problems previously that kept the dunbrack scoring machinery from properly evaluating the energies for terminal residues:

1. The code was using the sentinel value "-1" to represent too many things: in one step, that an atom defining a certain chi dihedral was not resolvable because, e.g., that atom would need to live inside residue i-1 for the very first residue or inside residue i+1 for the last residue; in another step, -1 represented that a residue did not exist or was not described by the dunbrack library (e.g. ALA or a ligand). The code was written to intercept the first case to define phi as the neutral -60 and psi as the neutral +60, but it never got the chance because the second case caused the residues to be skipped over entirely.
2. The code was trying to decide how to handle residue types based on their full names as opposed to their base names, so MET:nterm was being treaded as if it was not a residue type that Dunbrack had anything to say about (akin to ALA or a ligand)

The code now evaluates properly for the n-term of 1ubq

### DIFF
--- a/tmol/score/dunbrack/params.py
+++ b/tmol/score/dunbrack/params.py
@@ -706,6 +706,11 @@ class DunbrackParamResolver(ValidateAttrs):
         torch_device: torch.device,
     ) -> DunbrackParams:
         nstacks = res_names.shape[0]
+        # print("nstacks", nstacks)
+        # print("phi", phi.shape)
+        # print("psi", psi.shape)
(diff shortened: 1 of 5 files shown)


fdimaio (approved):
LGTM.  Maybe remove some of the commented debugging code in tmol/score/modules/dunbrack.py?

PR #269, opened 2023-08-30
Dimaio/bb dep omega

fdimaio:
This code adds a backbone-dependent omega term to tmol.  It matches the R3 implementation with 2 exceptions:
1) residues w/ undefined phi/psi are mapped to basic (non-bbdep) variant instead of setting undefined torsion to 0
2) preproline tables are not always used in R3 when they should be

The code replaces the rama and omega score terms with a single term, 'backbone_torsion,' that computes both in a single kernel.


tmol/tests/optimization/test_scorefunction_minimization.py
```diff
@@ -67,22 +64,17 @@ def run():
         cart_sfxn_network = CartesianSfxnNetwork(sfxn, pose_stack)
         optimizer = LBFGS_Armijo(cart_sfxn_network.parameters(), lr=0.1, max_iter=20)
 
-        # E0 =
         cart_sfxn_network.whole_pose_scoring_module(cart_sfxn_network.full_coords)
 
         def closure():
             optimizer.zero_grad()
             E = cart_sfxn_network()
+            print(E)
```
aleaverfay, line 72: ???

aleaverfay (changes requested):
Looks great! Just a few small questions and one suggestion w/ name-->base_name.

tmol/score/backbone_torsion/potentials/._compiled.ops.cpp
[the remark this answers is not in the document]
fdimaio: Sorry, this file shouldn't have been added (a temp file made by my editor)

tmol/score/backbone_torsion/potentials/backbone_torsion_pose_score.impl.hh
```diff
@@ -0,0 +1,261 @@
...
+        // omega : [V, dVdphi, dVdpsi, dVdomega]
+        accumulate<Dev, Vec<Real, 3>>::add(
+            dV_dxyz[1][pose_ind][phi_ats[j]], common::get<1>(omega).row(j));
+        accumulate<Dev, Vec<Real, 3>>::add(
+            dV_dxyz[1][pose_ind][psi_ats[j]], common::get<2>(omega).row(j));
+        accumulate<Dev, Vec<Real, 3>>::add(
+            dV_dxyz[1][pose_ind][omega_ats[j]], common::get<3>(omega).row(j));
+      }
+    } else {
+      // if rama is undefined, fall back to old version
```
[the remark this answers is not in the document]
fdimaio: Addressing the last three comments here:
 - line 189 was incorrect, should be checking omega
 - line 171 IS correct .. the bbdep should only get triggered if phi and psi are valid
 - if omega is valid and phi/psi are NOT, it should enter the block at 241 which computes the "old" omega (no phi/psi dependent)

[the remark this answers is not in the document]
fdimaio: I will rename valid_torsion to valid_phipsi

tmol/score/backbone_torsion/potentials/potentials.hh
```diff
@@ -0,0 +1,189 @@
...
+    tie(sig, dsigdphipsi) =
+        tmol::numeric::bspline::ndspline<2, 3, D, Real, Int>::interpolate(
+            coeffs_sig, phipsi_idx);
+
+    // energy
+    Real baseline = std::log(1. / (6. * sqrt(2. * pi)))
+                    - std::log(1. / (sig * sqrt(2. * pi)));
+    Real offset = omegaang.V * 180.0 / pi - mu;
+    if (offset < -180.0) {
+      offset += 360.0;
```
aleaverfay, line 149: oh weird! So, let's say that for a given phi/psi, the mu is 170. Then when omega is 175 it gets pushed toward 170. That makes sense to me. BUT if omega is cis and -5, then it's going to get pushed towards -10?? Is there really symmetry between mu for cis and trans omegas??

tmol/score/bonded_atom.py
```diff
@@ -133,4 +133,6 @@ def bonded_path_length(
         cache=True,
     )
 
-    return csgraph.dijkstra(bond_graph, directed=False, unweighted=True, limit=limit)
+    return csgraph.dijkstra(
+        bond_graph.tocsr(), directed=False, unweighted=True, limit=limit
```
aleaverfay, line 137: Does this call to bond_graph.tocsr() produce a faster traversal?

tmol/score/lk_ball/potentials/lk_ball.hh
```diff
@@ -1178,7 +1178,7 @@ TMOL_DEVICE_FUNC void lk_ball_atom_derivs_full(
   Eigen::Matrix<Real, n_lk_ball_score_types, 1> dTdV_local;
 
   for (int i = 0; i < n_lk_ball_score_types; ++i) {
-    dTdV_local[i] = dTdV[block_pair_dat.pose_ind][i];
+    dTdV_local[i] = dTdV[i][block_pair_dat.pose_ind];
```
aleaverfay, line 1181: I'm just a little surprised this is showing up as a difference for the merge. This change has already been made in trunk.

tmol/tests/support/test_database_converters.py
```diff
@@ -32,7 +33,7 @@ def test_rama_table_read(rosetta_database, default_database):
     )
     r3_paa_dir = os.path.join(rosetta_database, "scoring/score_functions/P_AA_pp/")
 
-    r3_general, r3_prepro = parse_all_tables(
+    r3_general, r3_prepro = rewrite_rama_binary.parse_all_tables(
```
aleaverfay, line 36: What would it take to get these tests running on the testing server?

tmol/tests/types/test_torch.py
```diff
@@ -121,7 +121,7 @@ def test_array_validation(example):
             assert not spec.validate(v)
 
 
-invalid_dtypes = [numpy.dtype([("coord", float, 3), ("val", int)]), numpy.complex, "c8"]
+invalid_dtypes = [numpy.dtype([("coord", float, 3), ("val", int)]), complex, "c8"]
```
aleaverfay, line 124: I'm trying to puzzle out: What is triggering the need for this change? I don't see any updates to the env.yml file that might cause a numpy version change.

aleaverfay (approved):
:+1: I just have a few non-blocking questions for you, but this looks great.

tmol/score/backbone_torsion/potentials/potentials.hh
fdimaio: when omega is cis it falls into the loop at 121 - 123 (no bb-dependence), 125-179 is run on trans only.

tmol/score/bonded_atom.py
fdimaio: I think this conversion is happening automatically, but newer versions of sparse don't do this autoconversion so we make it explicit.  I actually have a branch where this dependency is removed, but didn't want to duplicate it here.

tmol/tests/support/test_database_converters.py
fdimaio: So all it needs is pyrosetta installed as a dependency.  Not sure if that is worth it?

tmol/tests/types/test_torch.py
fdimaio: Not needed on test server, but I did some runs with newer versions of numpy and it triggered an error.

tmol/score/backbone_torsion/potentials/potentials.hh
fdimaio: I'll add comments clarifying this.

everyday847:
Question purely for documentation purposes -- what are the places in R3 where prepro isn't used where it should be? (Maybe this could inspire some improvements in R3.)

PR #270, opened 2023-09-08
Res-centric dunbrack

jflat06:
A rework of dunbrack to the res-centric scheme.

This still uses the DunbrackParamResolver, though portions of it to annotate the block types.

In the C++, it compresses everything down into a single kernel launch per residue.


jflat06:
I've cleaned up most of the code and added some of the requests from the cartbonded PR to this one as well.

... which actually let me find a bug on the jagged test in cuda:

```
E       RuntimeError: CUDA error: an illegal memory access was encountered
E       CUDA kernel errors might be asynchronously reported at some other API call,so the stacktrace below might be incorrect.
E       For debugging consider passing CUDA_LAUNCH_BLOCKING=1.
```

I'll be working on debugging this next. Hopefully something obvious.

tmol/score/dunbrack/potentials/dunbrack_pose_score.impl.hh
```diff
@@ -0,0 +1,384 @@
...
+    TView<Vec<int64_t, 3>, 1, D> semirot_table_strides,  // n-semirot-tabset
+    TView<Vec<Real, 3>, 1, D> semirot_start,             // n-semirot-tabset
+    TView<Vec<Real, 3>, 1, D> semirot_step,              // n-semirot-tabset
+    TView<Vec<Real, 3>, 1, D> semirot_periodicity,       // n-semirot-tabset
+
+    TView<Int, 1, D> block_n_dihedrals,
+    TView<UnresolvedAtomID<Int>, 2, D> block_phi_uaids,
+    TView<UnresolvedAtomID<Int>, 2, D> block_psi_uaids,
+    TView<UnresolvedAtomID<Int>, 3, D>
+        block_chi_uaids,  // TODO: unused, can probably be removed
```
fdimaio, line 80: Andrew can comment on this.  In theory a chi could cross a residue boundary ... but the code would need to be rewritten for that purpose...

tmol/score/dunbrack/potentials/dunbrack_pose_score.impl.hh
```diff
@@ -0,0 +1,384 @@
...
+  assert(pose_stack_block_type.size(0) == n_poses);
+  assert(pose_stack_block_type.size(1) == max_n_blocks);
+
+  assert(pose_stack_inter_block_connections.size(0) == n_poses);
+  assert(pose_stack_inter_block_connections.size(1) == max_n_blocks);
+  assert(pose_stack_inter_block_connections.size(2) == max_n_conns);
+
+  assert(block_type_atom_downstream_of_conn.size(0) == n_block_types);
+  assert(block_type_atom_downstream_of_conn.size(1) == max_n_conns);
+  // assert(block_type_atom_downstream_of_conn.size(2) == ); TODO: what?
```
fdimaio, line 119: @aleaverfay ?

fdimaio (approved):
LGTM

tmol/score/dunbrack/potentials/dunbrack_pose_score.impl.hh
jflat06: Ah, I forgot I had a couple TODOs in here.

I made the note because all of the chis are encoded in the block_dih_uaids (arg right below this). I originally separated them out, but it was easier to work with existing code by pooling them together. I'll go ahead and remove this unless there's a reason to keep it.

Side note: would it be beneficial to stick an extra check in our PR testing that looks for TODOs?

tmol/score/dunbrack/potentials/dunbrack_pose_score.impl.hh
jflat06: From what I remember from the code, it seems to be the max n atoms of the residues. I don't think there's an easy way to get at that value here.

I tried setting up the asserts to be very verbose, as I think this also serves a dual purpose of 'documenting' the arguments so that it's clear what each dimension is representing.

tmol/score/dunbrack/potentials/dunbrack_pose_score.impl.hh
```diff
@@ -0,0 +1,379 @@
...
+
+    ) -> std::tuple<TPack<Real, 2, D>, TPack<Vec<Real, 3>, 3, D>> {
+  int const n_poses = coords.size(0);
+  int const n_block_types = block_n_dihedrals.size(0);
+  int const max_n_atoms = coords.size(1);
+  int const max_n_blocks = pose_stack_block_coord_offset.size(1);
+  int const max_n_dih = block_dih_uaids.size(1);
+  int const max_n_conns = pose_stack_inter_block_connections.size(2);
+
+  int const DIH_N_ATOMS = 4;  // TODO: is there a global for this?
```
jflat06, line 100: Is there a global constant for this? Obviously very simple, but I think it makes the code clearer to use a name rather than a magic number.

PR #271, opened 2023-09-11
Fix error in name3 of patched residues

fdimaio:
Minor bug where name3 was getting overwritten in patching process.


aleaverfay (approved):
LGTM

PR #272, opened 2023-09-13
Ref Energy Term

jflat06:
This implements the ref energy term in pure python with the beta_nov16 weights (weights are defined in the DB).


tmol/score/ref/ref_energy_term.py
```diff
@@ -0,0 +1,105 @@
...
+        self.ref_weights["LEU"] = 1.65735
+        self.ref_weights["LYS"] = -1.34026
+        self.ref_weights["MET"] = -1.64321
+        self.ref_weights["PHE"] = -1.45095
+        self.ref_weights["PRO"] = -0.09474
+        self.ref_weights["SER"] = -0.28969
+        self.ref_weights["THR"] = 1.15175
+        self.ref_weights["TRP"] = 2.64269
+        self.ref_weights["TYR"] = 2.26099
+        self.ref_weights["VAL"] = 0.58223
```
jflat06, line 50: Obviously temporary.

tmol/score/ref/ref_whole_pose_module.py
```diff
@@ -0,0 +1,38 @@
...
+        self.ref_weights = _p(ref_weights)
+
+        # self.global_params = _p(torch.stack(_t([global_params.K]), dim=1))
+
+    def forward(self, coords):
+        score = torch.index_select(
+            self.ref_weights, 0, self.pose_stack_block_types.flatten()
+        )
+        score = torch.reshape(score, (self.pose_stack_block_types.size(0), -1))
+        score = torch.sum(score, 1, keepdim=True)
```
jflat06, line 36: Is this the most efficient way of doing this? It sounds like possibly flatten/reshape just create views instead of creating a whole new object, so I _think_ it should be fast.

tmol/score/ref/ref_energy_term.py
jflat06: If anyone has a pointer of where to find the values in Rosetta3 I can work on the actual DB implementation.

tmol/score/ref/ref_whole_pose_module.py
```diff
@@ -0,0 +1,38 @@
...
+
+        return score
```
jflat06, line 38: Is this the most efficient way of doing this? It sounds like possibly flatten/reshape just create views instead of creating a whole new object, so I _think_ it should be fast.

tmol/database/default/scoring/ref.yaml
```diff
@@ -0,0 +1,21 @@
...
+  LEU : 1.2334
+  LYS : -0.873554
+  MET : -5.1227
+  PHE : -1.0644
+  PRO : -1.281
+  SER : -1.1772
+  THR : -1.425
+  TRP : 2.085
+  TYR : 3.035
+  VAL : 0.964136
```
jflat06, line 21: The "method_weights" line in the wts file I took this from has no explicit ordering. 

I have been comparing the per-residue energies with pyrosetta, and they don't seem to match these values. I had assumed they were alphabetical.

I've searched Rosetta for the code responsible for Ref and came up with https://github.com/RosettaCommons/main/blob/8e2068a4eb74cbb1a14335fab8254a5094600cae/source/src/core/energy_methods/ReferenceEnergy.cc#L76 (note the comment: "This is a terrible terrible terrible hack that will do for now.")

Is the ordering in the lookup table coming from https://github.com/RosettaCommons/main/blob/8e2068a4eb74cbb1a14335fab8254a5094600cae/source/src/core/chemical/AA.hh#L43 ?

Also, the function in ReferenceEnergy.cc looks quite a bit more complicated than a single lookup table. Are any of these other cases things we need to worry about right now?

@aleaverfay @fdimaio

jflat06: Presumably answered my own question by copying that ordering, as the score for ubq1 now matches rosetta.

So that just leaves the question of whether there's any more functionality from ReferenceEnergy.cc that may be missing.

tmol/score/ref/ref_energy_term.py
```diff
@@ -0,0 +1,67 @@
...
+        return tmol.score.terms.ref_creator.RefTermCreator.score_types()
+
+    def n_bodies(self):
+        return 1
+
+    def setup_block_type(self, block_type: RefinedResidueType):
+        super(RefEnergyTerm, self).setup_block_type(block_type)
+
+        print(block_type.name3)
+        ref_weight = getattr(self.ref_weights, block_type.name3)
```
jflat06, line 38: Is name3 appropriate here?

tmol/score/ref/ref_whole_pose_module.py
```diff
@@ -0,0 +1,46 @@
...
+            self.ref_weights, 0, block_types[real_blocks]
+        )
+
+        # for each pose, sum up the block scores
+        score = torch.sum(score, 1)
+
+        # wrap this all in an extra dim (the output expects an outer dim to separate sub-terms)
+        score = torch.unsqueeze(score, 0)
+
+        score.requires_grad = True  # a bit of a hack to make the benchmark test not error out because there are no grads
```
jflat06, line 44: There's some question about the best way to handle this. We could also change the benchmarks to simply not try to run the backward pass explicitly for terms that don't have grads in their output.

tmol/database/scoring/ref.py
```diff
@@ -0,0 +1,39 @@
...
+    LEU: float
+    LYS: float
+    MET: float
+    PHE: float
+    PRO: float
+    SER: float
+    THR: float
+    TRP: float
+    TYR: float
+    VAL: float
```
fdimaio, line 28: I don't think we want to hardcode this.  In theory a user should be able to specify ref weights for other types, as long as base_name matches.  Just loading into a string->value dict should work.

tmol/score/ref/ref_whole_pose_module.py
fdimaio: I think this is fine

fdimaio (approved):
See my one comment but LGTM

tmol/score/ref/ref_energy_term.py
jflat06: Ok, I think github may be acting up for me or something. I was looking at this PR yesterday and did not see any comments from Frank, but now I see this comment from 2 days ago (along with some others).

I'll go ahead and make the change for adding the extra entries for HIS_D/CYD and changing it to base_name.

tmol/score/ref/ref_energy_term.py
```diff
@@ -0,0 +1,71 @@
...
+
+    def setup_block_type(self, block_type: RefinedResidueType):
+        super(RefEnergyTerm, self).setup_block_type(block_type)
+
+        if hasattr(block_type, "ref_weight"):
+            return
+
+        ref_weight = self.ref_weights[
+            block_type.base_name
+        ]  # TODO: 0.0 or throw error if key not found?
```
jflat06, line 39: @fdimaio 

What should be the behavior here? Default to zero or throw an exception?

PR #273, opened 2023-09-20
OOF. Out of bounds read in APSP. :pray: Karolis

aleaverfay:
It is shockingly important that you do not read from memory that is not yours.


fdimaio (approved):
LGTM.  Good catch

PR #274, opened 2023-10-02
Aleaverfay/fast cpu apsp

aleaverfay:
This PR speeds up the Floyd Warshall O(N^3) all-pairs-shortest-path algorithm In PoseStack construction with Dijkstra's single-source shortest-path-to-all-other-vertices O(N^2 + NKlnK) algorithm when a threshold/cutoff weight/distance is given where K is the expected number of nodes that will be traversed from each starting node before the threshold is reached. For an average sized antibody sequence (e.g. Pertuzumab) this is a roughly 100x speedup. The N^2 component is to simply initialize the output weight tensor / examine all input edge weights.


tmol/tests/pack/sim_anneal/test_sim_anneal.py
```diff
@@ -240,6 +240,7 @@ def dont_test_mc_accept_reject_module_smoke(ubq_res, default_database, torch_dev
     # print("alternate_ids", alternate_ids.shape)
     # print("faux_energies", faux_energies.shape)
 
+    print("calling mc_accept_reject go")
```
fdimaio, line 243: is this meant to be added?

tmol/tests/pose/compiled/test_all_pairs_shortest_paths.py
```diff
@@ -94,8 +100,44 @@ def test_all_pairs_shortest_paths_w_off_diagonal_bonds():
...
+    weights[:, arange_n_nodes[1:], arange_n_nodes[:-1]] = 1
+    weights[:, arange_n_nodes[:-1], arange_n_nodes[1:]] = 1
+    weights[:, 10, 20] = 1
+    weights[:, 20, 10] = 1
+
+    weights_cuda = weights.clone().to(torch_device_cuda)
+
+    stacked_apsp(weights, 6)
+    stacked_apsp(weights_cuda, 6)
+    weights_cuda[weights_cuda > 6] = 6
```
fdimaio, line 140: presumably cuda 'stacked_apsp' does not use this max count.  I think line 140 should be included in the cuda implementation ... we want this function to behave identically on CPU and CUDA.

tmol/utility/cpp_extension.py
```diff
@@ -21,8 +22,12 @@
 _default_include_paths = list(tmol_include_paths() + extern_include_paths())
 
 _required_flags = ["--std=c++14", "-DWITH_NVTX", "-w"]
-_default_flags = ["-O3"]
-# _default_flags = ["-g", "-Og", "-DDEBUG"]
+
+if os.environ.get("DEBUG"):
+    _default_flags = ["-O3", "-DDEBUG"]
```
fdimaio, line 27: do you also want to enable this on test server?  if so I think you need to change file in .buildkite

fdimaio (approved):
lgtm.  See comments

tmol/utility/cpp_extension.py
aleaverfay: Right!

PR #275, opened 2023-10-10
Aleaverfay/pose stack to canonical form

aleaverfay:
Add code to convert back from PoseStack format to the "canonical form" that's input to "pose_stack_from_canonical_form."

### DIFF
--- a/tmol/io/pose_stack_deconstruction.py
+++ b/tmol/io/pose_stack_deconstruction.py
@@ -1,0 +1,101 @@
+import torch
+
+from tmol.pose.pose_stack import PoseStack
+from tmol.io.canonical_ordering import (
+    ordered_canonical_aa_types,
+    ordered_canonical_aa_atoms_v2,
+    max_n_canonical_atoms,
+)
+from tmol.io.details.select_from_canonical import (
+    _annotate_packed_block_types_w_canonical_res_order,
+    _annotate_packed_block_types_w_dslf_conn_inds,
+    _annotate_packed_block_types_w_canonical_atom_order,
+)
+
+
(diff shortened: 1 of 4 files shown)


tmol/io/pose_stack_deconstruction.py
```diff
@@ -0,0 +1,245 @@
...
+        cf_coords,
+        atom_is_present,
+        disulfides,
+        res_not_connected,
+    )
+
+
+def _annotate_packed_block_types_w_termini_types(pbt: PackedBlockTypes):
+    # TEMP!
+    # for now: a bt is a down/up terminus if it includes
```
fdimaio, line 122: how is this going to be changed in the future? (why is it TEMP)?

fdimaio (approved):
Looks good.  A lot of the error handling code is not in codecov, but that is probably fine.

tmol/io/pose_stack_deconstruction.py
aleaverfay: The code I'm working on to generalize the termini patches looks at all the Patches in the PatchedChemicalDatabase and notes which patches remove the "{<down>}" or "{<up>}" atoms. It can then determine if a given patched block type is a termini variant by seeing what variant-display-names the variant's name includes. Right now, instead of using that code, I have effectively hard coded "there are only two types of patches that change the termini of residues and they are the "nterm" and "cterm" patches."

aleaverfay:
I'm really not sure why code-cov is highlighting the fact that there's some amount of untested code in the canonical-form-to-pose-stack code path here in this branch that's exclusively dedicated to the reverse pose-stack-to-canonical-form code path. The new code is all being tested, fwiw.

PR #279, opened 2023-10-12
Aleaverfay/create atom index mapping b/w PoseStack and CanonicalForm

aleaverfay:
Adding an extra argument to the pose_stack_from_canonical_form function, return_atom_mapping, that will cause the function to return a mapping between the atom indices in the input canonical form and the atom indices in the pose stack that was created. This mapping will ease the access into and out of PoseStack.

From the documentation:
```
return_atom_mapping: return the mapping for atoms in the canonical-form tensor
        to their PoseStack index; this could be used to update the coordinates
        in a PoseStack without rebuilding it (as long as the chemical identity
        is meant to be unchanged) or to perhaps remap derivatives to or from
        pose stack ordering. If requested, the atom mapping will be the last two
        arguments returned by this function, as two tensors:
            ps, t1, t2 = pose_stack_from_canonical_form(
                ...,
                return_atom_mapping=True
            )
            can_ord_coords[
                t1[:, 0], t1[:, 1], t1[:, 2]
            ] = ps.coords[
                t2[:, 0], t2[:, 1]
            ]
        where t1 is a tensor nats x 3 where
        - position [i, 0] is the pose index
        - position [i, 1] is the residue index, and
        - position [i, 2] is the canonical-ordering atom index
        and t2 is a tensor nats x 2 where
        - position [i, 0] is the pose index, and
        - position [i, 1] is the pose-ordered atom index
```

### DIFF
--- a/tmol/io/details/select_from_canonical.py
+++ b/tmol/io/details/select_from_canonical.py
@@ -297,17 +297,16 @@ def take_block_type_atoms_from_canonical(
     missing_atoms = torch.zeros(
         (n_poses, max_n_blocks, pbt.max_n_atoms), dtype=torch.int32, device=pbt.device
     )
+    real_canonical_atom_inds = canonical_atom_inds[real_atoms]
     block_coords[real_atoms] = coords[
-        nz_real_pose_ind, nz_real_block_ind, canonical_atom_inds[real_atoms]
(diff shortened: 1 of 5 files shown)


fdimaio (approved):
lgtm

PR #282, opened 2023-10-20
Remove old scoring code

fdimaio:
Some issues that have come up so far (will be updated as progress continues):
* There is no kinematic optimization module for new score system
* Much of the kinematic code testing used the old framework and was eliminated
* some other tests used old framework: hbond_identification, dunbrack_chi_sampler

With the latter two it is not clear if these tests are redundant or not.  We can look at code coverage to see.

I will next focus on adding a kinematic optimization module


jflat06:
You should be good to remove the old cartbonded database code: 

https://github.com/uw-ipd/tmol/blob/0a705c3fa797b0d3ff1258539a3034e8691cff3a/tmol/database/scoring/cartbonded.py#L8-L68

https://github.com/uw-ipd/tmol/blob/0a705c3fa797b0d3ff1258539a3034e8691cff3a/tmol/database/scoring/__init__.py#L4

https://github.com/uw-ipd/tmol/blob/0a705c3fa797b0d3ff1258539a3034e8691cff3a/tmol/database/scoring/__init__.py#L19

https://github.com/uw-ipd/tmol/blob/0a705c3fa797b0d3ff1258539a3034e8691cff3a/tmol/database/scoring/__init__.py#L34-L36

jflat06:
The DunbrackParamResolver:

https://github.com/uw-ipd/tmol/blob/0a705c3fa797b0d3ff1258539a3034e8691cff3a/tmol/score/dunbrack/params.py#L143

Is used by the new dunbrack code, but only a fairly small section of it.

You can see the scope of its usage by my new code here:

https://github.com/uw-ipd/tmol/blob/0a705c3fa797b0d3ff1258539a3034e8691cff3a/tmol/score/dunbrack/dunbrack_energy_term.py#L46-L51

It isn't immediately clear to me how to extricate the dead code from the ParamResolver, Since the new code calls from_database, which does a bunch of preprocessing of the database, it isn't as simple as removing unused functions. But I'm pretty sure most of this code is unnecessary now.

jflat06:
Can we remove all of omega, now that it included in a sub-term of another term? Or is that in another PR? Maybe you have already done this, and I'm reading the diff wrong.

jflat06:
https://github.com/uw-ipd/tmol/blob/0a705c3fa797b0d3ff1258539a3034e8691cff3a/tmol/database/default/scoring/omega.yaml 

https://github.com/uw-ipd/tmol/blob/0a705c3fa797b0d3ff1258539a3034e8691cff3a/tmol/database/scoring/__init__.py#L47

Looks like it can be removed, since you added new versions for the BBdependent version.

jflat06:
Can remove: https://github.com/uw-ipd/tmol/blob/0a705c3fa797b0d3ff1258539a3034e8691cff3a/tmol/database/default/scoring/cartbonded.old.yaml along with the other code from cartbonded above.

jflat06:
No code removed from tmol/score/hbond/params.py - other terms seemed to either cut this entirely or partially.

fdimaio:
Thanks! Will make these updates.

fdimaio:
@jflat06 OK this should be ready for review

aleaverfay:
This seems fine. It makes me think we ought to come up with some PoseStack-based initialization of the KinForest before we remove the PackedSystem code because we don't want to leave it untested or it will rot; so perhaps getting the PoseStack-based KinForest construction code is a higher priority than I first thought.

jflat06:
I'll give my 'approval' on the parts of the code I am familiar with, but there is lots left that I don't feel confident in approving.

I don't know if we want to handle the trimming down of the DunbrackParamResolver, but it looks like that's still untouched.

aleaverfay (approved):
Seems totally fine. I'm sure there is plenty more code we will want to remove, but there is no reason not to move forward with what you have already removed.

PR #285, opened 2023-12-05
Fix disulfide term's handling of placeholder residues (i.e. block type == -1)

aleaverfay:
Add new unit test for jagged poses which contain placeholder residues which failed in the absence of the if (block_type != -1) check


fdimaio (approved):
LGTM

PR #288, opened 2024-02-12
bump python/torch and update packaging

kleinhenz:
This enables tmol to work with the latest torch (2.2) and python (3.12). It also (hopefully) simplifies the packaging of tmol so that it can be used more easily in a greater number of environments.

* To support newer pytorch versions the compiler std version is bumped to c++17. There are also some additions of explicit device placements for tensors because torch has gotten more strict about automatically moving tensors in indexing operations. The deprecated `torch.testing.assert_allclose` is replaced by `torch.testing.assert_close` (see https://github.com/pytorch/pytorch/issues/61844)
* To support python >= 3.11 typish is removed as a dependency and replaced by a vendored version which fixes a bug in equality comparison of subscripted types.
* package configuration is moved into `pyproject.toml` and now includes specification of dependencies so that they are automatically fetched on install. This also enables installing directly from the git repo (` pip install git+https://github.com/kleinhenz/tmol.git@bump_versions`)
* git version information is embedded into the library and available at `tmol.__version__` using [setuptools_scm](https://github.com/pypa/setuptools_scm).
* The Dockerfile is simplified by using [micromamba base images](https://micromamba-docker.readthedocs.io/en/latest/)
* in the `env.yml` the cudatoolkit is changed to the version provided by nvidia at nvidia/label/cuda-12.1.1::cuda. All other dependencies are moved to pip since it makes propagating dependencies to other projects easier.
* A frozen set of known good dependency versions and a script to update them are provided in environments/linux-cuda.


aleaverfay (approved):
Merging to temporary intermediate for testing purposes

PR #291, opened 2024-03-19
RF2 Integration

jflat06:
This is a quick working example of getting RF2 and tmol talking.

I've tested an example in RF2 with minimization happening at each step, and putting the result back into the RF2 xyz. 

Small detail - the N terminus hydrogens and the C terminus OXT are not preserved if I fetch the minimization result from tmol and then send it back and rescore it. I assume this is expected?


tmol/__init__.py
```diff
@@ -26,6 +27,12 @@
     write_pose_stack_pdb,
     atom_records_from_pose_stack,  # TO DO: SHOULD THIS BE IN THE API??
 )
+
+from tmol.optimization.sfxn_modules import CartesianSfxnNetwork as cart_sfxn_network
+
+from tmol.optimization.lbfgs_armijo import LBFGS_Armijo as lbfgs_armijo
```
jflat06, line 33: Better way of doing these?

tmol/io/pose_stack_from_rosettafold2.py
```diff
@@ -130,6 +130,80 @@ def canonical_form_from_rosettafold2(seq, xyz, chainlens):
     )
 
 
+def pose_stack_to_rosettafold2(seq, xyz, chainlens, pose_stack):
```
jflat06, line 133: This name no longer really fits the file. How should we rename the function/file?

tmol/optimization/modules.py
```diff
@@ -74,6 +75,7 @@ def torsional_energy_network_from_system(
     )
 
 
+"""
```
jflat06, line 78: Commented this out with the above, not sure if technically necessary since I don't think it would have been parsed unless used.

tmol/io/pose_stack_from_rosettafold2.py
```diff
@@ -130,6 +130,80 @@ def canonical_form_from_rosettafold2(seq, xyz, chainlens):
...
+    canonical_form = canonical_form_from_pose_stack(co, pose_stack)
+
+    seq = seq.unsqueeze(0)
+    xyz = xyz.unsqueeze(0)
+    tmol_restypes = rf22t_rtmap[seq]  # reverse this mapping
+    atom_mapping = rf22t_atmap[seq]
+    rf2_at_is_real = rf2_at_is_real_map[seq]
+
+    rf2_coords = torch.full(  # allocate xyz for RF2 instead, with correct size - length (sum(L_s)), 27, 3
+        xyz.shape,
```
jflat06, line 192: Right now I just steal the shape from the original xyz that we pass in. I guess technically all it should need is the L_s and seq?

jflat06:
Closing in favor of https://github.com/uw-ipd/tmol/pull/298

PR #293, opened 2024-04-03
Dimaio/minimizer perf

fdimaio:
Fix the issue where DOF masking was leading to perf issues.

### DIFF
--- a/tmol/optimization/modules.py
+++ b/tmol/optimization/modules.py
@@ -14,40 +14,24 @@ from tmol.system.score_support import kincoords_to_coords
 #  - or we might want to keep that with dof creation
 
 
-# mask out relevant dofs to the minimizer
-class DOFMaskingFunc(torch.autograd.Function):
-    @staticmethod
-    def forward(ctx, fg, mask, bg):
-        ctx.mask = mask
-        ctx.fg = fg
-        bg[mask] = fg
-        return bg
-
-    @staticmethod
-    def backward(ctx, grad_output):
-        grad = torch.zeros_like(ctx.fg)
-        grad = grad_output[ctx.mask]
-        return grad, None, None
-
-
 # cartesian space minimization
 class CartesianEnergyNetwork(torch.nn.Module):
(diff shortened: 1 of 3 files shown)


tmol/optimization/modules.py
```diff
@@ -82,23 +66,29 @@ def __init__(self, score_system, dofs, kinforest, system_size, dof_mask=None):
...
 
+        if dof_mask is None:
+            dof_mask = torch.full(
+                dofs.shape[:-1], True, device=dofs.device, dtype=torch.bool
+            )
+
+        self.system_size = system_size
+
+        if dof_mask is None:
+            dof_mask = torch.full(
```
aleaverfay, line 77: This is unreachable / duplicated. I think it's safe to remove.

tmol/optimization/sfxn_modules.py
```diff
@@ -14,20 +13,23 @@ def __init__(
...
+        self.masked_coords = torch.nn.Parameter(self.full_coords[self.coord_mask])
         self.count = 0
 
     def forward(self):
         self.count += 1
-        self.full_coords = DOFMaskingFunc.apply(
-            self.masked_coords, self.coord_mask, self.full_coords
-        )
+        self.full_coords = self.full_coords.detach()
+        self.full_coords[self.coord_mask] = self.masked_coords
```
aleaverfay, line 32: So was the performance issue related to this scatter but that `.detach()` above resolves it?

aleaverfay (approved):
LGTM. Minor points in the comments.

tmol/optimization/modules.py
fdimaio: will do

tmol/optimization/sfxn_modules.py
fdimaio: there is a little overhead with the scatter (the 0.55s v 0.65s difference we currently see) but most of the overhead was from the memory allocation

PR #294, opened 2024-04-10
Dimaio/remove old scoreterms

fdimaio:
Rebasing this branch on block_pair_scoring branch.

Removes much of the code from the old scoring system.

See #282 for additional info.


jflat06:
Seems like codecov may have caught some more unused old scoring code that can be removed.

tmol/tests/data/__init__.py
```diff
@@ -76,53 +75,25 @@ def disulfide_res():
...
-        40: read_pdb(pdb.data["bysize_040_res_5uoi.pdb"]),
-        75: read_pdb(pdb.data["bysize_075_res_2mtq.pdb"]),
-        150: read_pdb(pdb.data["bysize_150_res_5yzf.pdb"]),
-        300: read_pdb(pdb.data["bysize_300_res_6f8b.pdb"]),
-        600: read_pdb(pdb.data["bysize_600_res_5m4a.pdb"]),
+        40: pdb.data["bysize_040_res_5uoi.pdb"],
+        75: pdb.data["bysize_075_res_2mtq.pdb"],
+        150: pdb.data["bysize_150_res_5yzf.pdb"],
+        300: pdb.data["bysize_300_res_6f8b.pdb"],
+        600: pdb.data["bysize_600_res_5m4a.pdb"],
```
jflat06, line 93: Interesting - do we want to use these at all for the score term test?

jflat06 (approved):
Seems good if the tests pass.

jflat06:
Just as a note, with our squashing the history, there's currently a *lot* of functionality wrapped up in a single commit with the block pair scoring PR now.

fdimaio:
> Just as a note, with our squashing the history, there's currently a _lot_ of functionality wrapped up in a single commit with the block pair scoring PR now.

This is true.  Perhaps it makes more sense to merge this branch first (what else is needed?) and then merge the code removal to master?

tmol/tests/data/__init__.py
fdimaio: I don't know that we need to.  This is used currently to show runtime scaling as a function of protein size.

jflat06:
> > Just as a note, with our squashing the history, there's currently a _lot_ of functionality wrapped up in a single commit with the block pair scoring PR now.
> 
> This is true. Perhaps it makes more sense to merge this branch first (what else is needed?) and then merge the code removal to master?

By 'this' branch do you mean the block_pair_scoring branch? I think that would be reasonable, but I am not hung up on it if we just want to combine them. More a note for the future that we may want to keep things more thematically cohesive (I am also guilty of this with including the testing infrastructure changes in with block_pair_scoring).

tmol/tests/data/__init__.py
jflat06: I was thinking of using these as a more realistic set of data for the 'jagged' score term test, but I guess different terms may need different inputs to ensure they're tested properly (disulfide, for example).

tmol/tests/score/test_score_function_benchmarks.py
```diff
@@ -236,7 +236,7 @@ def test_render_module(
     @benchmark
     def setup():
         sfxn = beta2016_score_function(torch_device)
-        scorer = sfxn.render_whole_pose_scoring_module(pose_stack)
+        _ = sfxn.render_whole_pose_scoring_module(pose_stack)
```
jflat06, line 239: Same as above - do we want to check anything with it afterward?

tmol/tests/score/test_score_function_benchmarks.py
```diff
@@ -216,7 +216,7 @@ def setup():
         canonical_form = canonical_form_from_pdb(
             co, systems_bysize[system_size], torch_device
         )
-        pose_stack = pose_stack_from_canonical_form(co, pbt, **canonical_form)
+        _ = pose_stack_from_canonical_form(co, pbt, **canonical_form)
```
jflat06, line 219: Do we want to check anything with it afterward, or is this just a smoke?

jflat06 (approved):
Seems good, still waiting on tests to finish to merge the block pair branch.

PR #298, opened 2024-04-17
RF2 Integration

jflat06:
Adds a pose_stack to RF2 function, tests for this function, and hooks for accessing some tmol functions from RF2.

### DIFF
--- a/tmol/__init__.py
+++ b/tmol/__init__.py
@@ -21,6 +21,7 @@ from tmol.io.pose_stack_from_openfold import (  # noqa: F401
 )
 from tmol.io.pose_stack_from_rosettafold2 import (  # noqa: F401
     pose_stack_from_rosettafold2,
+    pose_stack_to_rosettafold2,
     canonical_form_from_rosettafold2,
     canonical_ordering_for_rosettafold2,
     packed_block_types_for_rosettafold2,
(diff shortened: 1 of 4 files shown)


kierandidi:
@jflat06 should we close this PR since RF2 is out of date by now?

jflat06:
By RF2 out of date, do you mean that nobody uses RF2 now? Or that RF2 has changed in some way such as to not be compatible with this?

kierandidi:
> By RF2 out of date, do you mean that nobody uses RF2 now? Or that RF2 has changed in some way such as to not be compatible with this?

I mean that folks use RF3 and other models now

fdimaio:
only reason to maybe keep this is this is super easy to plug rf3 into as well (rf3 uses this format for conf head)

kierandidi:
Audited against current master/#465. The reverse conversion here hard-codes n_poses=1 and requires external chain lengths, while current tmol now has a batched atomworks_from_pose_stack round trip for the 37-slot model representation and #465 adds the prepared Atom37/Biotite path for modern model outputs and general chemistry. This branch is conflicting and its RF2-specific implementation has no current CPU/CUDA/PyTorch validation. I recommend closing it as superseded; if RF3 needs an exact confidence-head serialization contract, add that as a thin tested adapter over the current AtomWorks mapping rather than reviving this RF2-only path.

PR #299, opened 2024-06-03
Aleaverfay/kinforest from stencil

aleaverfay:
Create KinForests on the GPU (mostly).

The user specifies the kinematic system they want in terms of edges in a FoldForest:
- polymer edges for connections through the up & down connection points in polymeric residues (blocks)
- jump edges to connect arbitrary pairs of residues (using their default jump connection atoms), and
- root jump edges that connect a residue by a jump to the explicit virtual root at the origin

For each block type (residue type) with n_conn connections, the kinforest-from-stencil algorithm creates (n_conn + 2)^2 annotations as an (n_conn + 2) x (n_conn + 2) matrixing of input-connection sources x first-output-connection sinks. In addition to the n_conn connection point entries/first-exits, there are two additional input sources: jumps and root jumps; and there are two additional first output sinks: jumps and "no exit" (when the block is a leaf in the fold forest). For each of these combinations of input source and output sink, the algorithm annotates the block type with a set of _scan-path segments_ (SPSs) representing how the generational-segmented-scan algorithm will compute the coordinates for the atoms in a block of this block type in the context of the full structure; the assembly of the KinForest will join SPSs together to create _scan paths_ (SPs) that will calculate coordinates in the same segmented-scan generation.

The KinForest construction algorithm begins on the CPU with a depth-first traversal of the edges in the fold forest, then a leaf-to-root traversal to select the "first descendant" edges for each branching node (by minimizing the number of generations in the generational segmented scan algorithm), a root-to-leaf traversal to assign "edge delays" (how many generations will be traversed before any SPSs for the blocks on that edge are folded), and a topological sort to construct a total-ordering of the edges based on a partial ordering of 3-tuples of a) edge delay, b) breadth-first traversal order and c) the position of the edge in the unbreaking path of first descendants that will build it. Thus if the polymer edge (0, 10) has as its first descendant the jump (10, 100) and this edge has as its first descendant the polymer edge (100,200), then these three edges would end up "adjacent" in the topological sort. Thus we will put the atoms for residue 100 next to the atoms for residue 10 when constructing the scan path that begins at residue 0.

At this point, the computation moves to the GPU. From here, the GPU shuffles the SPS stencils for each block into the right position in the "nodes" tensor. To compute where each SPS is to be written, the algorithm must first compute offsets into the nodes tensor for each SPS by running scan on the number of atoms each SPS contains. The position where each SPS must write down its number of atoms in the number-of-atoms-for-SPS tensor is given by an offset based on the number of SPSs each residue has (which will be computed using scan), and the position where each residue should write down the number of SPSs it has (per generation) is given by an offset based on the number of residues that each edge builds (again computed using scan) where the edges write down their counts in the topological-sort order. Thus there are three scan steps to compute offsets. Offsets for offsets for offsets.

This PR removes the remnants of the old PackedResidueSystem which was preserved until now only to keep the unit tests running on the kinematics code; now that this algorithm is fully operational, the PackedResidueSystem can be removed.


tmol/chemical/restypes.py
```diff
@@ -259,6 +262,75 @@ def _setup_ordered_torsions(self):
...
 
+    @property
+    def n_torsions(self):
+        return self.ordered_torsions.shape[0]
+
+    is_torsion_mc: numpy.ndarray = attr.ib()
+
+    @is_torsion_mc.default
+    def _setup_is_torsion_mc(self):
+        # A torsion is a "main chain" torsion if all of its atoms are
```
fdimaio, line 273: Mainchain atoms are a property now?

What exactly depends on this?  is it just atom tree building?

tmol/extern/moderngpu/scan_types.hxx
```diff
@@ -0,0 +1,15 @@
...
+#include "meta.hxx"
+
+BEGIN_MGPU_NAMESPACE
+
+// Types for scan operations that are CPU-compatible.
+
+enum scan_type_t {
+  scan_type_exc,
+  scan_type_inc
+};
```
fdimaio, line 13: nice!

tmol/kinematics/fold_forest.py
```diff
@@ -17,36 +17,70 @@ class FoldForest:
...
+    connected to this root by a "root jump". Such root-jump residues are defined
+    by listing the residue that the root is connected to as the "end" residue;
+    the "start" residue field should be left as -1. An example FoldForest for a
+    ten-residue protein might be:
+      (polymer, 0, 4)
+      (jump   , 0, 7)
+      (polymer, 7, 9)
+      (polymer, 7, 6)
+      (root-jump, -1, 0)
+      (root-jump, -1, 5)
```
fdimaio, line 46: perfect

tmol/tests/optimization/test_kin_min.py
```diff
@@ -0,0 +1,47 @@
...
+    mm = MoveMap.from_pose_stack(pose_stack)
+    mm.move_all_jumps = True
+    mm.move_all_named_torsions = True
+
+    kf_sfxn_net = build_kinforest_network(pose_stack, sfxn, ff, mm)
+    assert kf_sfxn_net is not None
+
+
+def test_run_kin_min_smoke(
+    jagged_stack_of_465_res_ubqs: PoseStack,
```
fdimaio, line 30: so this tests with assorted fts in each layer?

fdimaio: In torsion space does this move bond angles/lengths or just torsions?

tmol/kinematics/move_map.py
```diff
@@ -0,0 +1,466 @@
...
+        max_n_atoms_per_block: int,
+        device: Optional[torch.device] = None,
+    ):
+        if device is None:
+            device = torch.device("cpu")
+
+        self.move_all_jumps = False
+        self.move_all_root_jumps = False
+        self.move_all_mc = False
+        self.move_all_sc = False
```
fdimaio, line 295: how does this handle torsion only, versus torsion + bond + angle?  (what is default)

fdimaio (approved):
LGTM.  See comments (mostly questions) below

tmol/score/common/uaid_util.hh
```diff
@@ -66,6 +66,50 @@ TMOL_DEVICE_FUNC int resolve_atom_from_uaid(
...
+                                                        // other block
+    if (idx < 0) {
+      return {-1, -1};
+    }
+    return {other_block_index, idx};
+  } else {
+    return {-1, -1};
+  }
+}
+
```
jflat06, line 112: Possible to re-use any code from the above function for this?

jflat06 (commented):
A lot here that is beyond my current understanding. If you want, we could meet up and go over the details (though I have a feeling this would take a while). 

There's a few suggestions and one data member name typo fix that will otherwise cause a bug.

tmol/score/common/uaid_util.hh
aleaverfay: Yes, indeed, the duplication is unnecessary.

aleaverfay (commented):
Thank you for the feedback! I left comments and addressed the suggestions that you both made.

PR #301, opened 2024-07-03
Return dictionary instead of tuple from canonical_form_from_pose_stack

aleaverfay:
The proper "canonical form" is a dictionary of the arguments that are input into pose_stack_from_canonical_form and thus a dictionary should be returned from the function in pose_stack_deconstrction.py that turns a pose_stack into a canonical form.


fdimaio (approved):
LGTM

PR #302, opened 2024-07-10
Activate the missing `ref` energy in the beta2016_score_function

aleaverfay:
The ref term was overlooked in the function that returns tmol's version of the beta2016 score term. Oops! Let's go ahead an activate it now.


fdimaio (approved):
lgtm

PR #305, opened 2024-07-23
Constraints

jflat06:
You can create a constraint set and add constraints to it with:

add_constraints(your_constraint_function, atom_indices, parameters)

atom_indices: a tensor that is (constraint_ind x atom_num x (pose_ind, res_ind, atom_ind). You must supply at least 2 and at most 4 atoms in the atom_num dimension. 

parameters: a (constraint_ind x 6) float tensor that you can use to store parameters on a per-constraint basis. 

your_constraint_function: this must take two args, (atoms, params). atoms will be filled with a (n_constraints x 4 x 3) tensor containing atom coordinates. If you supplied less than 4 atom indices when you created the constraints, you must not use any extra atoms (they will be filled with garbage). parameters will be the parameter tensor you supplied when you added the constraints.

You can call add_constraints multiple times and they will be concatenated.

An example:

```
    def constfn(atoms, params):
        return params[:, 0]

    cnstr_atoms = torch.full((2, 2, 3), 0, dtype=torch.int32, device=torch_device)
    cnstr_params = torch.full((2, 6), 0, dtype=torch.float32, device=torch_device)

    cnstr_atoms[0,0] = torch.tensor([0,0,0]) # constraint 0, atom 0 - assigned to atom at pose 0, res 0, atom 0
    cnstr_atoms[0,1] = torch.tensor([0,1,1]) # constraint 0, atom 1 - assigned to atom at pose 0, res 1, atom 1
    cnstr_params[0, 0] = 2 # constraint 0, parameter 0 - assigned to 2

    cnstr_atoms[1,0] = torch.tensor([1,0,0])
    cnstr_atoms[1,1] = torch.tensor([1,1,1])
    cnstr_params[1, 0] = 4

    constraints.add_constraints(constfn, cnstr_atoms, cnstr_params)

    cnstr_atoms[0,0] = torch.tensor([0,1,0])
    cnstr_atoms[0,1] = torch.tensor([0,2,1])
    cnstr_params[0, 0] = 20

    cnstr_atoms[1,0] = torch.tensor([1,0,0])
    cnstr_atoms[1,1] = torch.tensor([1,2,1])
    cnstr_params[1, 0] = 40

    constraints.add_constraints(constfn, cnstr_atoms, cnstr_params)
```

This returns:

```
[[[[ 0.  1.  0.  0.  0.]
   [ 1.  0. 10.  0.  0.]
   [ 0. 10.  0.  0.  0.]
   [ 0.  0.  0.  0.  0.]
   [ 0.  0.  0.  0.  0.]]

  [[ 0.  2. 20.  0.  0.]
   [ 2.  0.  0.  0.  0.]
   [20.  0.  0.  0.  0.]
   [ 0.  0.  0.  0.  0.]
   [ 0.  0.  0.  0.  0.]]]]
```

Right now, the score for a constraint is statically split between the residues of the first and second atoms. You could utilize this convention to manipulate where the scores end up by ordering the atoms in your input atom index tensor appropriately. I'm not sure if that's ideal, but may be OK for now.


tmol/score/constraint/constraint_energy_term.py
```diff
@@ -0,0 +1,142 @@
...
+        self.device = device
+
+    @classmethod
+    def score_types(cls):
+        import tmol.score.terms.constraint_creator
+
+        return tmol.score.terms.constraint_creator.ConstraintTermCreator.score_types()
+
+    def n_bodies(self):
+        return 1
```
jflat06, line 33: Not totally sure what to do here, since this could involve up to 4 'bodies'. But as I understand it, this is mostly for dispatch reasons, which isn't a concern in this case.

tmol/score/constraint/constraint_energy_term.py
```diff
@@ -0,0 +1,142 @@
...
+        # g1 = torch.logical_and(lb <= dist, dist <= ub) # default 0
+        g2 = torch.logical_and(ub < dist, dist <= ub2)
+        g3 = dist > ub2
+
+        ret[g0] = ((lb - dist) / sd) ** 2
+        # ret[g1] = 0
+        ret[g2] = ((dist - ub) / sd) ** 2
+        ret[g3] = 2 * rswitch * ((dist - ub) / sd) - rswitch**2
+
+        return ret
```
jflat06, line 71: This logic could use some checking (or a baseline comparison vs Rosetta)

tmol/score/constraint/constraint_energy_term.py
```diff
@@ -0,0 +1,142 @@
...
+    @classmethod
+    def circularharmonic_angle(cls, atoms, params):
+        return cls.circularharmonic(atoms, params)
+
+    @classmethod
+    def circularharmonic_torsion(cls, atoms, params):
+        return cls.circularharmonic(atoms, params, torsion=True)
+
+    @classmethod
+    def circularharmonic(cls, atoms, params, torsion=False):
```
jflat06, line 82: Is there a cleaner way of breaking this up?

tmol/score/constraint/constraint_whole_pose_module.py
```diff
@@ -0,0 +1,112 @@
...
+                + atom_res_inds[:, 0] * nblocks
+                + atom_res_inds[:, 1]
+            )
+            indices2 = (
+                atom_pose_inds[:, 0] * (nblocks**2)
+                + atom_res_inds[:, 1] * nblocks
+                + atom_res_inds[:, 0]
+            )
+            flattened.index_add_(0, indices1, scores / 2)
+            flattened.index_add_(0, indices2, scores / 2)
```
jflat06, line 91: Right now this splits the score between the residues of the first and second atoms that were provided.

This makes it a bit awkward because it means that a torsion has to be specified in (1)(4)(2)(3) order rather than (1)(2)(3)(4) order if you want to have the score split between them.

I could change it to be the 1st and 4th atoms instead, but this would mean that someone writing a simple distance constraint would have to provide a tensor with 4 atoms (and then only use the 1st and 4th in their function). I'm not sure what the best solution is to this.

tmol/score/score_types.py
```diff
@@ -7,6 +7,7 @@ class ScoreType(AutoNumber):
     cart_torsions = ()
     cart_impropers = ()
     cart_hxltorsions = ()
+    constraint = ()
```
jflat06, line 10: All under a single score term right now.

tmol/tests/score/constraint/test_constraint_energy_term.py
```diff
@@ -0,0 +1,187 @@
...
+    cnstr_params = torch.full((1, 4), 0, dtype=torch.float32, device=torch_device)
+
+    res1_type = pose_stack.block_type(0, 0)
+    res2_type = pose_stack.block_type(0, 1)
+    cnstr_atoms[0, 0] = torch.tensor([0, 4, res1_type.atom_to_idx["C"]])
+    cnstr_atoms[0, 1] = torch.tensor([0, 5, res2_type.atom_to_idx["N"]])
+    cnstr_params[0, 0] = 1.47  # lb TODO
+    cnstr_params[0, 1] = 1.47  # ub TODO
+    cnstr_params[0, 2] = 1.47  # sd TODO
+    cnstr_params[0, 3] = 1.47  # rswitch TODO
```
jflat06, line 45: Some examples of a realistic bounded constraint would be very appreciated.

tmol/pose/constraint_set.py
```diff
@@ -0,0 +1,115 @@
...
+            (0, 1), 0, dtype=torch.float32, device=device
+        )
+        self.device = device
+
+    #################### PROPERTIES #####################
+
+    def add_constraints_to_all_poses(self, fn, nposes, atom_indices, params=None):
+        self.add_constraints(fn, atom_indices, params, nposes=nposes)
+
+    def add_constraints(self, fn, atom_indices, params=None, nposes=0):
```
jflat06, line 38: Right now adding constraints to all poses is done with a different function. This was originally done by simply omitting the 'pose' index from the atom specifiers. I'm not sure which is preferable.

jflat06, line 38: We have to pass in the number of poses in order to replicate a copy of each constraint to all poses, since this constraint_set doesn't have a reference to the pose. 

I'm thinking more and more that these functions should be going through pose stack (and then maybe managed internally in the pose_stack by a constraint_set class?). The constraints are fundamentally attached to the pose_stack anyway, especially if you get the atom names from the pose stack's block types like in the test cases.

jflat06:
@fdimaio 

This should be ready for review. I've tagged a number of notes/questions in the code for some feedback.

Also, I took a look at the tests failing and it seems like something might be up with the testing server:

⚠️ Warning: Checkout failed! Error running `/usr/bin/git remote set-url origin https://github.com/uw-ipd/tmol`: exit status 128 (Attempt 3/3 Retrying in 2s)
🚨 Error: Error running `/usr/bin/git remote set-url origin https://github.com/uw-ipd/tmol`: exit status 128

tmol/score/constraint/constraint_energy_term.py
jflat06: Remove the non-torsion version.

tmol/score/constraint/constraint_whole_pose_module.py
jflat06: Copy last atom into slot 4.

tmol/score/score_types.py
jflat06: Save this for later.

tmol/tests/score/constraint/test_constraint_energy_term.py
fdimaio: something like 1,3,1,1

tmol/pose/constraint_set.py
fdimaio: CSTs live in posestack
Deep copy on pose copy
None if no csts

fdimaio: See above

tmol/score/constraint/constraint_energy_term.py
jflat06: Look at electrostatics for an example of sliding atom for testing.

fdimaio (approved):
lgtm (pending tests)

PR #306, opened 2024-11-20
use header only nvtx3

kleinhenz:
With the most recent pytorch release (2.5.1) I've seen some linking failures related to nvtx: `undefined symbol: nvtxRangePushA`. This seems to be because `libtorch_cuda` no longer links `libnvToolsExt.so`. Luckily [nvtx](https://github.com/NVIDIA/NVTX) is now header only so the resolution is simple.


aleaverfay (approved):
LGTM; waiting on results from testing server.

PR #309, opened 2025-02-26
Dunbrack bin refactor

jflat06:


aleaverfay:
What version of torch introduced "safe_globals"? I had previously been using torch 2.2 in one of those environments but I get an error saying that the function doesn't exist.

PR #310, opened 2025-02-27
Update ATen dispatch calls to use Tensor::options() rather than Tensor::type()

fdimaio:
This enables compatibility w/ torch2.6

Note: we do not update the testing env.


jflat06 (approved):
Seems good. Is this backward compatible with older versions of torch?

fdimaio:
> Seems good. Is this backward compatible with older versions of torch?

It should be (the test server will show if there are problems).

aleaverfay (approved):
LGTM!

PR #311, opened 2025-03-06
Aleaverfay/small fixes

aleaverfay:
Fix a handful of warnings and a compilation issue.

Also add a much-needed function to `ScoreFunction` that lists the currently active set of `ScoreType`s that the `ScoreFunction` is using so that we can iterate across the tensor returned by `WholePoseScoringModule::unweighted_scores` and report which term has which weight.


fdimaio (approved):
lgtm

PR #312, opened 2025-03-07
Add coordinate constraint support

aleaverfay:
If you want to idealize a structure, one of the necessary components is to be able to constrain to particular coordinates. Coordinate constraint support is added here by mimicking the calculations for harmonic distance restraints except that the target distance between the actual and the target coordinates, x0, is fixed at 0.

Adding a unit test to show how to add coordinate constraints from a reference PDB of identical chemical composition for heavy atoms and a PDB of a beta2016-minimized copy of 1ubq whose coordinates have moved from the starting coordinates. I've tested but not committed a version of kinematic minimization using coordinate constraints and it successfully pulls this kin-minimized ubq back toward the starting coordinates (though not all the way).

We should probably allow a spring-constant parameter, k, for both coordinate constraints and distance restraints.


kierandidi:

aleaverfay:
> @aleaverfay do we still want to merge this or shoud we close this PR?

We should merge this. I'll resolve the conflicts but then I'd appreciate a review

fdimaio (approved):
LGTM

PR #313, opened 2025-03-07
First attempt at resetting R3 count-pair style i to i+2 connectivity

aleaverfay:
A little hacked together piece of code to set the i-to-i+2 chemical bond separation count to have C of residue i and N of residue i+2 be treated at full strength to match Rosetta3's count pair "logic"


kierandidi:

kierandidi:
Current-code audit: this prototype is unsafe to merge. The optional hack slices global nonzero residue lists and assumes adjacent entries are i/i+2 in the same pose and chain, so batch boundaries, chain breaks, padding, and non-polymeric residues can be paired incorrectly. It is restricted to alpha amino acids, contains unconditional debug prints, and has no behavioral tests. If Rosetta3 count-pair compatibility is still needed, it should be implemented as a named count-pair policy using explicit per-pose connectivity and validated with golden pair weights across termini, breaks, multiple chains, and batches.

PR #315, opened 2025-04-30
Torch 2.6 and removing Zarr

jflat06:
Updating to torch 2.6 
Removing zarr and replacing previous zarr-based DB loading with torch.load

Remade this branch since I git merged in the previous one, and that messes badly with codecov results.


fdimaio (approved):
lgtm

PR #316, opened 2025-05-07
C++/CUDA pre-built binaries

jflat06:
Adding code to pre-build all the these modules as a distributable wheel.

Changing C++ interface from torch to pybind.


kierandidi:
@jflat06 can we close this PR?

PR #318, opened 2025-06-04
Fix simulated annealer test

jflat06:


tmol/utility/autograd.hh
jflat06, line 1: Only referenced in an unused kinematics source file and this packer code. Was not actually used at all, and was also failing to compile.

jflat06:
Not needed anymore

PR #319, opened 2025-06-20
bugfix for kintree stencil creation

aleaverfay:
Fixing a very small issue in the creation of scan orderings for leaf nodes in the FoldForest.

When processing the leaf-node exit-path assignment, (ie. j == n_conn +1), then whether an exit-path is available as one of the descendants does not matter, and thus it is usually preferable for CA to choose CB as its first child so it can build the side chain in the first path rather than e.g. C, where it would dead-end.


kierandidi:

fdimaio (approved):
LGTM

PR #321, opened 2025-08-28
Rotamer centric scoring

jflat06:


jflat06:
@aleaverfay Are we good to remove all of the ljlk/potentials/rotamer_pair_energy_* files? I don't believe they're used, but I want to make sure we don't have plans for them before I remove them.

jflat06:
> @aleaverfay Are we good to remove all of the ljlk/potentials/rotamer_pair_energy_* files? I don't believe they're used, but I want to make sure we don't have plans for them before I remove them.

score/common/count_pair.hh included both lj.hh and rotamer_pair_energy_lj.hh, but neither of these appears to be used.

tmol/score/energy_term.py
```diff
@@ -84,5 +94,54 @@ def setup_poses(self, pose_stack: PoseStack):
...
+
+    def get_rotamer_score_term_function(self):
+        raise NotImplementedError()
+
     def render_whole_pose_scoring_module(self, pose_stack: PoseStack):
-        pass
+        try:
+            f = self.get_pose_score_term_function()
+        except NotImplementedError:
+            f = self.get_score_term_function()
```
jflat06: @aleaverfay what is the purpose of blocks like this? Are we trying to account for a case where there are not separate pose and block scoring functions? Does/should that happen?

tmol/tests/data/term_baselines/CartBondedEnergyTerm/test_whole_pose_scoring_10.yaml
```diff
@@ -1,55 +1,55 @@
...
-  pose0: 50.58418273925781
-  pose1: 50.58419418334961
-  pose2: 50.58417892456055
-  pose3: 50.584205627441406
-  pose4: 50.58418655395508
-  pose5: 50.584197998046875
-  pose6: 50.584190368652344
-  pose7: 50.58419418334961
-  pose8: 50.58420944213867
-  pose9: 50.58420181274414
```
jflat06: Large difference.

jflat06:
TODO: assure assert blocks are filled out in the energy terms

tmol/tests/data/term_baselines/CartBondedEnergyTerm/test_block_scoring.yaml
```diff
@@ -1,85 +1,85 @@
 term0:
   pose0:
     res0:
-      res0: 0.4605046510696411
-      res1: 0.0
```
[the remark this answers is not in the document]
aleaverfay: Intended :+1:

tmol/tests/data/term_baselines/CartBondedEnergyTerm/test_whole_pose_scoring_10.yaml
aleaverfay: Same as above

PR #348, opened 2026-03-16
Utility to add coordinate constraints for the mainchain atoms 

aleaverfay:
Fast coordinate constraint addition using the residue-params->properties->polymer->mainchain_atoms list and creating an annotation in the PackedBlockTypes tensor


aleaverfay:
Covered in #350

PR #349, opened 2026-03-16
Aleaverfay/dlin2

aleaverfay:
Making the distance factor of lj-sigma at which the repulsive component goes linear (the lj_dlin_sigma_factor) configurable so that we can soften the lj-rep component programmatically rather than having it hard coded.


tmol/score/lk_ball/potentials/gen_pose_waters.impl.hh
```diff
@@ -223,6 +223,7 @@ struct GeneratePoseWaters {
     int const n_blocks = n_poses * max_n_blocks;
     DeviceOps<Dev>::template foreach_workgroup<launch_t>(n_rots, f_watergen);
 
+    DeviceOps<Dev>::synchronize_device();
```
fdimaio, line 229: I think this should be removed?

fdimaio (approved):
This looks fine to me, though I think the device synch should be removed (not sure who added this?)

PR #350, opened 2026-03-20
Aleaverfay/relax w coord csts

aleaverfay:
Add option to Relax to add coordinate constraints to the mainchain atoms to anchor them to the starting coordinates.


fdimaio (approved):
LGTM

PR #354, opened 2026-03-26
Localized Packer

jflat06:
This adds an alternative to the existing simulated annealer that focuses each CTA on an individual residue, with the threads in that CTA iterating through the trajectories, and for each trajectory, splitting the work of considering different rotamers between the threads in their group.

Results seem to match the existing anneal (and curiously, beating it in some cases...?), but the performance is currently much worse.

This seems to be some issue with the process of selecting a rotamer from the 32 considered rotamers in the CTA. Time to pack a sidechainless 1ubq goes from 0.05s to 1.24s when commenting in/out the selection code.


jflat06:
Archiving this for now

PR #357, opened 2026-04-05
Support for Apple Silicon (Metal/MPS)

fnachon:
This is an MPS fork I needed for a Proteina-Complexa fork running on Apple Silicon.

Florian


kierandidi:
This is not merge-ready against the current build matrix. Concrete blockers found in the audit: it accidentally renames/deletes tmol/kinematics/compiled/compiled_inverse_kin.py into a .cmake query file; commits .claude/settings.local.json, CMake query artifacts, and a generated tmol_primitives.metallib; changes the package repository URL to the contributor fork; enables MPS based on runtime hardware availability rather than an explicit tested build target; and has no current macOS MPS build/runtime lane. It also conflicts with the current CPU/CUDA and PyTorch 2.13/2.14 C++ standard/toolchain work. This needs a clean, narrowly scoped successor with generated artifacts excluded and three independent lanes: Linux CPU regression, Linux CUDA regression, and macOS arm64 MPS compile+forward/backward parity.

PR #363, opened 2026-04-14
Aleaverfay/bump check

aleaverfay:
This examines the energies of the rotamers after they have all been computed to remove rotamers that are very unlikely to be selected in the final rotamer assignment because their total energy is greater than 5 kcal/mol under even the most favorable rotamer assignment of its neighbors. This eliminates about 1/3rd of the rotamers that are originally built. This should give a modest performance boost to packing.

I'll do some measurements.


kierandidi:
@fdimaio seems like this is ready for review by you? Just closing up some old PRs, I think we still want to merge this one @aleaverfay ?

tmol/pack/packer_task.py
```diff
@@ -398,6 +399,33 @@ def disable_packing_by_block_mask(self, block_type_mask: Tensor[torch.bool][:, :
...
+        with each neighbor across all the neighbor's rotamers.
+
+        bump_check removes ~40% of all rotamers and can significantly
+        improve running time, but, this comes at the expense of
+        eliminating rotamers that are sometimes the best option
+        when all others are bad in ways that bump-check's rosie
+        estimation cannot predict. In ~10% of tested crystal
+        structures packed with only the base rotamers (no ex flags),
+        bump_check increased the energy of the final rotamer
+        rotamer assignment by >20 kcal/mol.
```
fdimaio, line 419: Andrew ... is this a concern?  that seems ... quite high to me?

tmol/pack/compiled/compiled.impl.hh
```diff
@@ -52,70 +72,757 @@ auto InteractionGraphBuilder<DeviceDispatch, D, Real, Int>::f(
...
+      if (block1 > block2) {
+        printf(
+            "Assumption violated! block1 %d < block2 %d for index %d (pose "
+            "%d)\n",
+            block1,
+            block2,
+            index,
+            rot1,
+            rot2);
+      }
```
fdimaio, line 143: i think this will crash? should be (roti %d rotj %d)?

tmol/pack/compiled/compiled.impl.hh
```diff
@@ -1,4 +1,5 @@
 #pragma once
+#include <iostream>
```
fdimaio: still needed? i dont know if we want this?

fdimaio (commented):
OK, so the _code_ is fine (modulo comments below).  BUT:
```
In ~10% of tested crystal
        structures packed with only the base rotamers (no ex flags),
        bump_check increased the energy of the final rotamer
        rotamer assignment by >20 kcal/mol.
```
makes me a little concerned about making this on by default in relax?

Can we inspect some of these cases?

tmol/pack/packer_task.py
aleaverfay: It's a huge concern! It definitely is worth testing whether this has negative consequences in relax, but, fwiw, the structure of relax somewhat ensures that bump check will not make a huge difference: in particular, by starting at a very low fa_rep weight, the percentage of rotamers eliminated by bump check in the initial packing is very low, and then at higher fa_rep weights, the current rotamer usually wins, and so the likelihood that some high-energy rotamer will be eliminated despite being the best of a bad lot is much lower. I think we should do this: turn it off in relax by default, merge this branch, then benchmark to make sure that enabling it in relax does not worsen its performance.

fdimaio: I like this plan.

fdimaio (approved):
LGTM (pending removing on-by-default in relax)

PR #366, opened 2026-04-17
Dimaio/opt h flip hnq

fdimaio:
Adds an optH option to tmol.

Implemented through an OptHSampler that works with the existing build_rotamers sampler, sampling proton positions for C/T/Y/S as specified in parameters.  Takes a single option (default True) fix_HNQ that additionally considers flips of H/N/Q as well as tautamerization changes of H.

The model is turn ON by default in pose_stack_from_biotite, but can be disabled with no_optH=True.  pose_stack_from_openfold/rosettafold2 are unchanged.

New test added, small bug in fixed_aa_chi_sampler fixed.


tmol/pack/rotamer/fixed_aa_chi_sampler.py
```diff
@@ -56,7 +56,6 @@ def sample_chi_for_poses(
                 for one_pose_blts in task.blts
                 for blt in one_pose_blts
                 for i, bt in enumerate(blt.considered_block_types)
-                if self in blt.conformer_samplers
```
aleaverfay, line 59: Could you please explain why this line is removed? If a given position does not include the FixedAAChiSampler, it seems like it should not go ahead and build them at this position, right?

tmol/pack/build_missing_sidechains.py
```diff
@@ -6,61 +6,96 @@
...
+    pose_stack: PoseStack,
+    sfxn: ScoreFunction,
+    dunbrack_sampler: DunbrackChiSampler,
+    block_has_missing_atoms: Tensor[torch.bool][:, :],
+    rts,
+) -> PoseStack:
+    """Backward-compatible wrapper around build_missing_sidechains.
+
+    Calls build_missing_sidechains with no_optH=True, preserving the original
+    behavior: Dunbrack for blocks with missing heavy atoms, all others frozen.
```
aleaverfay: I'm not sure we want to preserve the original behavior. It's not like there's a lot of code out there reliant on this function

tmol/relax/fast_relax.py
aleaverfay: Okay, so, there now is separately the idea of including current and having an IncludeCurrentSampler. Does that mean that if you use one and not the other then you will not end up including the current rotamer? Can we have just one?

tmol/tests/pack/test_pack_rotamers.py
```diff
@@ -121,40 +71,136 @@ def test_pack_rotamers(default_database, ubq_pdb, dun_sampler, torch_device):
...
+        n_rots = int(rotamer_set.n_rots_for_block[pose_i, block_j].item())
+        rot_off = int(rotamer_set.rot_offset_for_block[pose_i, block_j].item())
+        for r in range(n_rots):
+            co = int(rotamer_set.coord_offset_for_rot[rot_off + r].item())
+            rc4 = rotamer_set.coords[[co + int(a4[k]) for k in range(4)]].double()
+            rot_chi = float(_cd(rc4[0:1], rc4[1:2], rc4[2:3], rc4[3:4])[0])
+            delta = math.degrees(rot_chi - input_chi)
+            # normalise to (-180, 180]
+            delta = (delta + 180.0) % 360.0 - 180.0
+            print(
```
aleaverfay: Can this be made an assertion?

tmol/pack/rotamer/opth_sampler.py
```diff
@@ -0,0 +1,578 @@
...
+    - HIS/HIS_D: {HIS, HIS_D} x {current chi2, chi2+180} = 4 rotamers.
+      All atoms through CG are taken from the input; ring atoms are rebuilt
+      from ideal geometry for three non-input variants.
+
+    NOTE: DunbrackChiSampler and OptHSampler must not be assigned to the
+    same block (Dunbrack already samples proton chis, so both on one block
+    oversamples). Assigning them to different blocks in the same task is fine.
+    """
+
+    fix_NHQ: bool = True
```
aleaverfay: "Fix" has three possible meanings:
1. hold still / prevent from changing
2. clean up by flipping
3. clean up by both flipping and changing the coords to ideal
The old name for this behavior was "flip_HNQ" and you use it in several places, but, not here?

tmol/pack/rotamer/opth_sampler.py
```diff
@@ -0,0 +1,578 @@
...
+        n_samples_per_chi=numpy.zeros(0, dtype=numpy.int32),
+        nhq_chi_col=nhq_chi_col,
+        nhq_chi_atom=nhq_chi_atom,
+        nhq_chi_4atoms=nhq_chi_4atoms,
+        is_his=is_his,
+    )
+
+
+def _downstream_kfo_atoms(rt) -> list:
+    """Return KFO indices of all atoms downstream of rt's NHQ chi-defining atom.
```
aleaverfay, line 75: This feels like it should be an annotation on the RTs / PBTs rather than something that gets built every time this code executes

tmol/pack/rotamer/opth_sampler.py
```diff
@@ -0,0 +1,578 @@
...
+    reset_kto, reset_bt, reset_k = [], [], []
+    chi_kto, chi_val_list = [], []
+
+    flip_sis = (chi_atoms >= 0).any(dim=1).nonzero(as_tuple=True)[0].tolist()
+    for si in flip_sis:
+        bt_idx = bt_inds[si].item()
+        at_off = at_offs[si].item()
+
+        # Step 2: reset downstream atoms to ideal for NHQ flip rotamers
+        if fix_NHQ:
```
aleaverfay: Does this mean that it's not possible to consider the input coordinates of the HNQ, that both flipped and un-flipped states are idealized?

aleaverfay (approved):
LGTM 👍 

Couple of questions and small suggestions

tmol/pack/rotamer/fixed_aa_chi_sampler.py
fdimaio: Shoot i think that's a mistake.

tmol/pack/build_missing_sidechains.py
fdimaio: ~~I'll adjust the comment a bit but i think old behavior might be desired sometime?~~
Looking at the wrong thing; I'll remove this.

tmol/pack/rotamer/opth_sampler.py
fdimaio: i'll change universally to flip

tmol/tests/pack/test_pack_rotamers.py
fdimaio: will do.  asserted all deltas are 0 or 180.

tmol/relax/fast_relax.py
fdimaio: ~~old one was universal.  new one can be applied at block level.  I'll remove old one.~~

Actually it looks like there is an include current field in blt.  I'm leaning to keeping the IncludeCurrentSampler to be more consistent?  But are there reasons not to?  I know in R3 ordering of taskops was a concern... is that an issue here?

tmol/pack/rotamer/fixed_aa_chi_sampler.py
fdimaio: ~~I misunderstood how this worked (and made same mistake in optH sampler).  Both are fixed now.~~

No I think I did understand and it is not a mistake.  when multiple samplers are provided, merge_conformer_samples does an element-wise torch.add of every sampler's n_rots_for_gbt, so each sampler must return a tensor whose length equals the total number of (pose, blt, considered_bt) entries.

I moved this check further downstream (line 64, `(self in blt.conformer_samplers and bool(blt.block_type_allowed[i]))`) ..  the code properly returns 0 rotamers for this position rather than skipping the position altogether.

This was caught by our test suite.

tmol/relax/fast_relax.py
fdimaio: Current version uses `IncludeCurrentSampler` but i can revert if you would prefer.  I like the idea that the task op is a container for samplers only (and fallback logic).

tmol/pack/rotamer/opth_sampler.py
fdimaio: ~~only for the atoms downstream of the chi to flip.  Is this undesired?~~

OK, on closer inspection, 'flip_sis' does _not_ contain the original rotamer for HNQ.  

Code is a bit confusing here, I added some comments.  The key is, that block is only called for rotamers that have an "override" of their chi values, which rotamer 0 of NHQ do not.

tmol/pack/rotamer/opth_sampler.py
fdimaio: added to RT cache

PR #367, opened 2026-04-22
Dimaio/minimizer updates

fdimaio:
Grab bag of changes:
1) fix lbfgs bug I inadvertently introduced trying to make the update step faster.  Instead make the Hessian update a very small CUDA kernel.
2) updates to hbond: change coefficients from hardcoded double->'Real', and split the graph traversal and scoring passes into two separate dispatch steps.
Other changes to lbfgs to make it more "Rosetta-like".


tmol/optimization/lbfgs_armijo.py
```diff
@@ -2,6 +2,8 @@
 from functools import reduce
 from torch.optim import Optimizer
 
+from tmol.optimization.compiled import lbfgs_two_loop as _lbfgs_two_loop_op
```
aleaverfay: Let's confine our imports from the .compiled files to inside of functions to avoid the circular import issues

tmol/optimization/lbfgs_armijo.py
```diff
@@ -402,61 +389,97 @@ def linefn(alpha_test):
...
                 prev_loss,  # current function value (at x)
-                alpha0=t,  # stepsize
+                alpha0=start_t,  # stepsize
                 factor=0.5,
                 sigma_decrease=0.1,
                 sigma_increase=0.8,
                 minstep=self._minstep,
             )
 
+            if self.verbose:
```
aleaverfay: This is *very* verbose.

tmol/score/common/device_operations.hh
```diff
@@ -4,7 +4,7 @@
 
 #include <tmol/utility/tensor/TensorPack.h>
 #include <tmol/utility/tensor/TensorAccessor.h>
-#include <tmol/extern/moderngpu/scan_types.hxx>  // CPU-friendly
+#include <moderngpu/scan_types.hxx>  // CPU-friendly
```
aleaverfay, line 7: In non-cuda builds, the include path to tmol/extern had previously not ended up in the g++ command line and so I had to change this #include to start at the base of the tmol/ directory; has that changed?

aleaverfay (commented):
Still reading: some short thoughts.

tmol/optimization/lbfgs_armijo.py
fdimaio: useful for tracking a debug trajectory though?  I removed it

tmol/score/common/device_operations.hh
fdimaio: so i did not try.  but I was getting errors compiling this with nvcc+JIT ... we include using signature 'tmol/extern/moderngpu/scan_types.hxx' here but internally mgpu uses 'moderngpu/scan_types.hxx'.  In git build this was causing an error (duplicated headers) ... some googling showed pragma once is weirdly behaved if things are included from different base paths?

I will test w/ CPU and address.

tmol/optimization/lbfgs_armijo.py
fdimaio: ok.  i've done this here and globally (13 other places).

jflat06 (approved):
HBond code looks good to me.

tmol/score/common/device_operations.hh
fdimaio: ok i tested CPU-only and it does work with our current setup.

tmol/tests/data/term_baselines/HBondEnergyTerm/test_block_scoring.yaml
```diff
@@ -3,7 +3,7 @@ term0:
     res0:
       res0: 0.0
       res1: 0.0
-      res2: -1.2079193592071533
+      res2: -1.211342692375183
```
aleaverfay: Why does this change?

aleaverfay (changes requested):
I'm a bit concerned by the somewhat large changes to the hbond energy -- it seems well outside the range I'd expect from the double->float conversion.

aleaverfay (approved):
After merging the double->Real reversion, this branch looks good to go 👍

PR #369, opened 2026-04-24
revert double to real conversion of hbond polys

fdimaio:
Reversion of double->Real conversion of hbond polys (and reversion of unit test changes)

Runtime of this version:
```
  structure    nres   t_tmol(s)    t_ros(s)   E_tmol_start         E_tmol     E_tmol_ros    E_ros_start          E_ros
  ---------  ------  ----------  ----------  -------------  -------------  -------------  -------------  -------------
  1INP          400        7.36      146.55       1307.469      -1030.550       -618.331       1703.449       -806.548
```
Versus float:
```
  structure    nres   t_tmol(s)    t_ros(s)   E_tmol_start         E_tmol     E_tmol_ros    E_ros_start          E_ros
  ---------  ------  ----------  ----------  -------------  -------------  -------------  -------------  -------------
  1INP          400        6.31      110.70       1309.654       -985.013       -580.288       1703.449       -799.147
```


aleaverfay:
Could you clarify just a few things with this test?
1. Does the input 1INP structure contain hydrogens already, or is optH running on the input structure? There's a 2kcal/mol difference in the starting energy and it's hard to know whether that's because of differences in the energy evaluation or differences in the starting coordinates.
2. Is minimization taking the same number of steps in these two executions? The double-precision version is getting a -45 kcal/mol better energy compared to the single-precision version and that difference feels like it could be the result of the minimizer taking more successful steps.

I spent a little time looking into a Kahan-summation algorithm for Horner's rule (https://hal.science/hal-00107222/document) and think that it's unlikely that we would see improved runtime using ~50 flops per iteration where previously it was 2 double precision operations.

I *do* think, though, that we can save most polynomial evaluations by exiting early from the polynomial evaluation if any of the three geometries are out-of-range; we already do an A-H distance check, but we can also check AHD and BAH.

fdimaio:
1INP does not have H.  so unfortunately differences are combo of hbond and optH.

Minimization takes similar #s of steps in both cases (randomness of packing leads to +/- 5 iterations to converge and it's within that delta).  What you say is correct but I would not read so much into this one point, we probably need to larger-scale testing.  You can see "randomness" in Rosetta's energies varying by non-trivial amount in two runs.  I'd like to merge, though, and move on to some other things, so I am more than happy to keep the double precision.

Perhaps we keep double precision for now and do more extensive testing on single versus double?

> previously it was 2 double precision operations.

keep in mind it's like 11 coeffs * 3 (dist, cosAHD, BAH) so 33 double ops per eval.

aleaverfay (approved):
I agree: let's merge this in to the minimizer_updates branch and merge minimizer_updates into trunk. We can separately re-evaluate the double->Real transition in the context of a full relax benchmark.

The point regarding N=1 energy improvement is well taken. I am highly suspicious of being able to take a 10^-5 step size to minimize in an energy landscape where the noise is 10^-2, though.

To clarify, the EFTHorner algorithm is 47 flops per polynomial degree (6 for TwoSum, 17 for TwoProd; done twice for the two polynomials -- the regular polynomial and the error polynomial -- plus 1 final sum) rather than traditional Horner's rule of 1 multiply + 1 add (or perhaps just a single fused-multiply-add operation). For our 11th-degree polynomials (12 iterations), that's 564 flops rather than 24 double precision ops (1692 for the three AH/BAH/AHD polynomials vs 72). The key that I think makes double precision preferable is that most threads are looking at atom pairs that don't form hbonds, and so when there's only 4 (or 8?) double-precision ALUs per SM, and the threads need to queue up, only a tiny number of them actually get in the queue.

PR #371, opened 2026-04-28
Fix Block Pair Scorefunction

jflat06:
Gets the block pair score function working again and also adds options to both whole_pose and block_pair for getting per-term and/or unweighted results.


fdimaio (approved):
LGTM

PR #372, opened 2026-04-28
Pinned version of sympy conflicts with torch req in git ci

jflat06:


fdimaio:
I do not think sympy is needed anymore (at least not directly)?

jflat06:
> I do not think sympy is needed anymore (at least not directly)?

Yeah I dont think so either, but the current ci script is pinning it because the container has it, I think (and then the pinned version conflicts with torch). I'm trying to fix this and some other issues with the build step.

jflat06:
Closed in favor of https://github.com/uw-ipd/tmol/pull/373

PR #373, opened 2026-04-28
Better git ci apptainer containerization

jflat06:
* Better isolation into a container environment with a new SIF file.
* Now fail on pre-commit with black, clang, flake8.
* Generate and upload Codecov properly.


fdimaio (approved):
lgtm!

PR #375, opened 2026-05-01
Handle extraneous atoms

jflat06:
Modifying select_from_canonical to be more forgiving of extraneous atoms. Throw warnings when these (or missing) atoms are detected.


tmol/io/pose_stack_from_biotite.py
jflat06, line 227: @aleaverfay @fdimaio 

This code was supposed to prevent the inclusion of H on the nterm in the first place, but was broken.

I've fixed it, but now this PR now fixes this in two ways: 
1. They are filtered here.
2. select_from_canonical now also penalizes instead of excludes candidates if they aren't filtered.

This leaves the question of which solution is preferable. 

I'm leaning towards simply deleting this filtering because:

* The penalty in select_from_canonical also seems more general.
* The filtering performed in this function must match the filtering performed downstream in select_from_canonical for correctness, which I'm not sure it does, currently.

tmol/io/details/select_from_canonical.py
```diff
@@ -508,10 +516,11 @@ def select_best_block_type_candidate(
     )
     real_candidate_misalignment_score = (
         2 * real_candidate_n_canonical_atoms_not_provided
+        + 10 * real_candidate_n_extraneous_atoms_provided
```
jflat06: @fdimaio Does this weight seem ok for penalizing extra atoms? That would be 5x the penalty of a missing atom.

tmol/io/details/select_from_canonical.py
```diff
@@ -506,19 +514,19 @@ def select_best_block_type_candidate(
     real_candidate_n_canonical_atoms_not_provided = torch.sum(
         real_candidate_canonical_atom_was_not_provided, dim=1
     )
+    warning_threshold = 2 * (canonical_ordering.max_n_canonical_atoms + 1)
     real_candidate_misalignment_score = (
         2 * real_candidate_n_canonical_atoms_not_provided
+        + warning_threshold * real_candidate_n_extraneous_atoms_provided
```
jflat06, line 520: We're using the previous failure_score multiplied by the number of extraneous atoms. This means that candidates where atoms are only missing will be preferred first over candidates where atoms would be extraneous.

tmol/io/details/select_from_canonical.py
```diff
@@ -595,26 +602,40 @@ def select_best_block_type_candidate(
...
+                "failed to resolve a block type from the candidates available\n",
+                *err_msg,
+            ]
         )
 
+        failure_threshold = 2 * warning_threshold
+        if torch.any(best_candidate_score >= failure_threshold):
+            raise RuntimeError(err_msg + "Best candidate exceeds failure threshold")
+        else:
+            logger.warning(err_msg)
```
jflat06, line 637: If the score exceeded the threshold by a factor of 2, lets just actually fail.

We should only get here if there were 2+ extraneous atoms in the best candidate, since we penalize each one by warning_threshold.

@aleaverfay @fdimaio does this sound good?

tmol/tests/io/details/test_select_from_canonical.py
```diff
@@ -1089,11 +1089,17 @@ def test_select_best_block_type_candidate_error_impossible_combo(
...
+  atom OP2 missing but present in candidate SER:phospho
+  atom OP3 missing but present in candidate SER:phospho
  Failed to resolve block type for 0 19 SER
  0 19 2 76 SER:mospho restype 15 equiv class SER
   atom HG provided but absent from candidate SER:mospho
   atom P provided but absent from candidate SER:mospho
-"""
+  atom OM1 missing but present in candidate SER:mospho
+  atom OM2 missing but present in candidate SER:mospho
+  atom OM3 missing but present in candidate SER:mospho
```
jflat06, line 1101: I added the missing atoms to the error as well to be a bit more verbose. I can remove this if we want, though.

tmol/io/details/select_from_canonical.py
fdimaio: I think this should be higher perhaps?

fdimaio (approved):
LGTM

tmol/io/details/select_from_canonical.py
aleaverfay: That means we'll tolerate 1 extra atom and then beyond that, refuse? That will give us basically the functionality we have now except that it'll handle an "H" on the N-term where it would've previously refused. That sounds reasonable.

jflat06: Yes, with the current threshold it will accept 1 extraneous atom. We could modify this to accept whatever number of these we're willing to accept, if so desired. But it should always prefer candidates with less.

PR #379, opened 2026-05-14
Add code to split out a single Pose from a PoseStack

aleaverfay:
Add tests with it to make sure that we can split PoseStacks and preserve their content.

Also fix several issues with concatenating ConstraintSets!


tmol/pose/pose_stack.py
```diff
@@ -231,6 +238,44 @@ def clone(self) -> "PoseStack":
...
+                index : index + 1
+            ]
+            .detach()
+            .clone(),
+            inter_block_bondsep=self.inter_block_bondsep[index : index + 1]
+            .detach()
+            .clone(),
+            inter_block_bondsep64=self.inter_block_bondsep64[index : index + 1]
+            .detach()
+            .clone(),
```
fdimaio, line 265: the autoformatting looks pretty bad here...

fdimaio (approved):
This looks fine to me.  But is it possible to make posestack[i] call this functionality also (return  the i-th pose)

PR #383, opened 2026-06-04
Key on hash of cart-bonded params in PBT caching

aleaverfay:
Add sha1 hash to cartbonded parameter set so that different cartbonded parameter sets can be cached in the block types and PBTs.

### DIFF
--- a/tmol/score/cartbonded/cartbonded_energy_term.py
+++ b/tmol/score/cartbonded/cartbonded_energy_term.py
@@ -1,5 +1,6 @@
 import torch
 import numpy
+import attrs
 
 from itertools import permutations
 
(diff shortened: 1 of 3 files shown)


tmol/score/cartbonded/cartbonded_energy_term.py
```diff
@@ -335,15 +370,17 @@ def get_score_term_attributes(self, pose_stack):
         def _t(ts):
             return tuple(map(lambda t: t.to(torch.float), ts))
 
+        print("Cart bonded get_score_term_attributes:", self.hash)
```
fdimaio: remove

fdimaio (commented):
lgtm (remove print though)

fdimaio (approved):
lgtm

PR #385, opened 2026-06-12
Soft repulsive

jflat06:


tmol/score/lk_ball/potentials/gen_pose_waters.impl.hh
```diff
@@ -226,7 +226,6 @@ struct GeneratePoseWaters {
     DeviceOps<Dev>::template foreach_workgroup<launch_t>(
         mgr, n_rots, f_watergen);
 
-    DeviceOps<Dev>::synchronize_device();
```
jflat06, line 229: @aleaverfay I left this call in when I merged your branch, but I'm removing it here unless there's a reason not to.

https://github.com/uw-ipd/tmol/pull/349/changes/BASE..be0041735b4bff5bf0e8a6975d65aa4c583725e6#r3358884814

fdimaio (approved):
Couple comments:
1) beta_soft also changes scoreterm weights.  Is that encoded here somehow?  getting beta16 and setting 'soft' is NOT the same as beta_soft in Rosetta proper
2) can you add a test for this?
Preemptively approving but please do this (and ensure tests pass)

jflat06:
@fdimaio Do we want to mimic the Rosetta pattern of a .wts file with options encoded in it? That seems like the most straightforward way to provide different pre-packaged score functions.

fdimaio:
> @fdimaio Do we want to mimic the Rosetta pattern of a .wts file with options encoded in it? That seems like the most straightforward way to provide different pre-packaged score functions.

I think so.  Probably yaml files to be consistent with everything else.  Perhaps database/score_functions?

You can then change the default loaders (currently hardcoded) to reference these files

jflat06:
Intended test will use ddg, but that only exists in the ligand branch. Going to push this to master and then update ligand_clean and add the test there.

tmol/database/default/scoring/ref.yaml
```diff
@@ -21,3 +21,26 @@ weights:
...
+    MET: 2.1654
+    ASN: -0.7549
+    PRO: -4.6330
+    GLN: -0.1943
+    ARG: -0.3519
+    SER: -1.2864
+    THR: -1.2064
+    VAL: 3.5934
+    TRP: 5.8434
+    TYR: 2.7219
```
fdimaio: so it seems weird to me for this to be in a different place than the scorefunction as the reference weights are very much tied to a particular scorefunction.  is it possible to move this to live in score_functions yaml instead?

I'm not sure how that plumbing will work though

fdimaio (approved):
lgtm

tmol/database/default/scoring/ref.yaml
jflat06: I can think about a more graceful way of doing this.

jflat06: My original thought on this was that for LJLK, we went with the paradigm where the alternate value for soft is encoded in the database, and we're merely selecting it.

The analogous model in ref would be to encode the values in the database as well, with soft_rep just selecting them.

One could imagine an "override" system for parameters in the database as an alternative, but this is de facto just evolving the database, which we had decided against previously.

jflat06:
Subsumed by kdidi/ligand_clean

PR #388, opened 2026-06-22
Adding version field to .tmol files.

jflat06:
This is probably a good step unless we're absolutely certain that the file formats will never change. 

I made it so that it requires the version field, but we could have it assume 1.0 if the field is missing, if people have a preference. 

Also added a version field to the .sfxn files.

### DIFF
--- a/tmol/tests/torch.py
+++ b/tmol/tests/torch.py
@@ -6,7 +6,7 @@ import torch
 import torch.cuda
 
 requires_cuda = pytest.mark.skipif(
-    lambda: not torch.cuda.is_available(),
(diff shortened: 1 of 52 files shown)


tmol/tests/torch.py
```diff
@@ -6,7 +6,7 @@
 import torch.cuda
 
 requires_cuda = pytest.mark.skipif(
-    lambda: not torch.cuda.is_available(),
```
jflat06, line 9: This was causing all cuda tests to skip because it wasn't evaluating the lambda, it was just taking the value of the lambda (which is True for lambdas).

PR #393, opened 2026-07-08
Dimaio/lk weird c c correction

fdimaio:
2 fixes to make LK and lk_ball match R3 exactly:
1) short-range spline starts on nearest etable grid boundarie rather than exactly 0.89*lj_radius
2) C-C correction (_only_ CH1/2/3/Caro) increases short-range cutoff to 4.2A


tmol/score/ljlk/potentials/lk_isotropic.hh
```diff
@@ -110,14 +110,16 @@ struct lk_isotropic_pair {
...
 
-    Real cpoly_close_dmin = d_min * d_min - 1.45;
-    if (cpoly_close_dmin < 0.01) cpoly_close_dmin = 0.01;
-    cpoly_close_dmin = std::sqrt(cpoly_close_dmin);
-
-    Real cpoly_close_dmax = std::sqrt(d_min * d_min + 1.05);
+    // close-spline knots lie on etable bins spaced 1/20 A^2 apart
+    Real const n = std::floor(Real(20) * d_min * d_min);
+    Real cpoly_close_dmin = std::sqrt(std::max(Real(0), n - 29) / 20);
+    Real cpoly_close_dmax = std::sqrt(std::min(n + 21, Real(405)) / 20);
```
jflat06, line 122: Just checking if any of these constants aren't already used/defined somewhere else, and whether we could factor these out instead of being magic hard coded values.

jflat06 (approved):

PR #394, opened 2026-07-14
Add a reusable context to pose_stack_from_biotite

fdimaio:
Allows faster rescoring of multiple poses with the same ligand:
```
        context = build_context_from_biotite(struct0, dev, prepare_ligands=True)
        for struct in structures:
            pose_stack = pose_stack_from_biotite(struct, dev, context=context)
```

Adds caching of the default databases' RTSs so adding a ligand to default DB does not trigger expensive recomputation.

### DIFF
--- a/tmol/tests/chemical/test_residue.py
+++ b/tmol/tests/chemical/test_residue.py
@@ -1,3 +1,4 @@
+import attr
 import cattr
 import numpy
 
@@ -6,6 +7,7 @@ from tmol.chemical.restypes import (
     RefinedResidueType,
     ResidueTypeSet,
 )
+from tmol.database.chemical import RawResidueType
 
 
 def test_refined_residue_construction_smoke(default_database):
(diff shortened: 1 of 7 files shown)


tmol/tests/chemical/test_residue.py
```diff
@@ -95,6 +97,40 @@ def test_residue_type_set_get_default():
...
+    copy = cattr.structure(cattr.unstructure(original), RawResidueType)
+    assert copy is not original
+    extended = attr.evolve(chem, residues=(*chem.residues, copy))
+
+    default_ids = {id(rt) for rt in ResidueTypeSet.get_default().residue_types}
+    s = ResidueTypeSet.from_database(extended)
+
+    # standard residues still reuse the cached objects ...
+    for rt in s.residue_types[: len(chem.residues)]:
+        assert id(rt) in default_ids
```
jflat06: Would we want to check that they are the same object in memory with "is" here? Could it theoretically rebuild it with the same ids?

jflat06 (approved):

tmol/tests/chemical/test_residue.py
fdimaio: probably but because packing annotates residue types we actually have to do shallow copies now.

PR #396, opened 2026-07-14
Enable ligand proton chi sampling by default

kierandidi:
## Summary
- enable `sample_proton_chi=True` by default in both Biotite pose-building entry points
- support the Cartesian product of multiple ligand proton chis and populate their defining atoms
- order double-digit chi names numerically and add focused OptH/default regression coverage

## Known failing ligand and failure
This is intentionally a draft because one end-to-end gate is still failing.

- Fixture: `tmol/tests/data/protein_ligand_test/cif_inputs/ace.ligand.cif`
- Prepared residue: `LG1`
- Failing path: default-on `pose_stack_from_biotite(..., prepare_ligands=True)` after OptHSampler packing
- Failure: `_assert_no_nan_coords()` reports NaN coordinates for ligand polar hydrogens `HN1`, `HN2`, `HN3`, `HO1`, `HO2`, `HO3`, and `HO4`
- Diagnostic trace: input coordinates and measured input DOFs are finite, while residue-type ideal DOFs are NaN for the traced sampled hydrogens `HN1`, `HN2`, `HN3`, `HO2`, `HO3`, and `HO4`
- Control: explicit `sample_proton_chi=False` builds the same ACE/LG1 pose with finite coordinates

No other ligand fixture has been identified as failing yet; ACE/LG1 is the reproduced case currently covered by the end-to-end regression.

## Verification
- pre-commit `black` and `flake8`: pass
- focused default-forwarding, mixed-radix proton-chi, non-finite-DOF, and numeric chi-order tests: pass
- end-to-end ACE/LG1 default-on pose regression: **fails** with the NaN atoms listed above

## Next step
Repair ligand ideal internal-coordinate construction for sampled polar hydrogens, then require the ACE/LG1 end-to-end regression and CI to pass before marking this PR ready for review.


tmol/pack/rotamer/opth_sampler.py
```diff
@@ -177,16 +180,23 @@ def _opth_fill_dofs(
     real_mask = dummy < n_atoms_per_rot.unsqueeze(1)
     dst = (at_offs.unsqueeze(1).expand(-1, pbt.max_n_atoms) + dummy)[real_mask]
     src = (orig_at_offs.unsqueeze(1).expand(-1, pbt.max_n_atoms) + dummy)[real_mask]
-    conf_dofs_kto[dst + 1, :] = orig_dofs_kto[src + 1, :]
+    copied_dofs = orig_dofs_kto[src + 1, :]
+    nonfinite = ~torch.isfinite(copied_dofs).all(dim=1)
+    if torch.any(nonfinite):
```
fdimaio: i _really_ don't think we want this here

the pose coming into optH should already be complete and well formed.

for proteins, we use 'generate_missing_leaf_atoms' to build missing H subject to ideal geometry, _then_ call this function.  ligands should use same logic.  my guess is the logic in that function, not here, needs to get fixed.

tmol/pack/rotamer/opth_sampler.py
```diff
@@ -746,35 +756,60 @@ def _fill_proton_chi_for_all_blocks(
...
-        max_n_expanded_chi = opth_cache.expanded_samples.shape[1]
-        chi_for_rotamers[rotamers_w_proton_chi_samples, :max_n_expanded_chi] = (
-            opth_cache.expanded_samples[
-                bt_for_proton_rotamer, :, sample_ind_for_proton_rotamer
-            ]
+        n_chi_cols = min(
+            chi_for_rotamers.shape[1],
+            opth_cache.expanded_samples.shape[1],
         )
+        for chi_col in range(n_chi_cols):
```
fdimaio, line 777: this is completely fine and how R3 works.  But you might want to talk to @aleaverfay about this, one downside is with large ligands the number of rotamers blows up exponentially.  It should be possible to treat each proton chi independently in the packer.

That might be a future effort though, it's fine for now.

kierandidi: @aleaverfay any idea comment on this here?

tmol/pack/rotamer/opth_sampler.py
kierandidi: sounds good, seems like you changed that by now?

tmol/pack/rotamer/opth_sampler.py
fdimaio: we talked about it at the meeting yesterday, it's a to do but not in this PR

tmol/pack/rotamer/opth_sampler.py
fdimaio: yes fixed now

tmol/pack/rotamer/opth_sampler.py
aleaverfay: The combinatorics of ligand / sugar hydroxyls was a big motivating factor for talking about blocks rather than residues. "Residue" is a somewhat rigid idea but "block" has no particular connotation to anything biochemical: it's just a unit on which you operate. The plan for hydroxyls is that they will be broken off from the rest of the residues that they come from and treated as independent blocks. Each block will have its own set of rotamers. What we have to create to make that work is:

1. a way to split a residue into multiple blocks,
2. a way to update all of the parameters in the scoring layer to reflect this splitting,
3. a way to convert a PoseStack created with the un-split block types into one with split block types, and
4. a way to convert a PoseStack back to its un-split block types form to return structures back to the user in a format that they will understand and be willing to work with

It feels like a decent amount of work, but one limiting factor until now has been that we don't even have any non-amino-acid residue types to start working from, so your push on the ligands is a big step forward on getting this implemented

PR #401, opened 2026-07-24
Refactor ligand conformer generation code.

fdimaio:
Rewrite ligand conformer code (previously using openbabel WeightedRotorSearch), to use custom code similar in principle to RDKit's ETKDG.

Significantly faster than previous implementation.  Properly handles chirality, large rings, pentavalent phosphates.

Also cleans up ligand code:
- dead code removal
- 4 test-only modules moved to test module
- removed bond type inference (we do not want to support this)
- updated README.md


kierandidi (approved):
Looks great, will merge!

PR #404, opened 2026-08-07
Dimaio/dna types and potential

fdimaio:
Add RNA and DNA support

## chemical and nonbonded parameters

DNA parameters come from a variant of beta (with Cameron Glasscock) that works well on protein/DNA interfaces: unique DNA atom types with LJ/LK/hbond weights fit by optE. RNA reuses these almost verbatim -- nonbonded geometry is close enough that the atom types and charges carry over, and only the icoors differ slightly (Engh and Huber for both).

## na_torsion / na_torsion_well

Rosetta scores DNA and RNA torsions with two different potentials that capture the same underlying property. In tmol, we went with the functional form of Rosetta's dna_torsion potential for both polymers, with only parameters differing between RNA and DNA.

The potential is rotameric -- each torsion has 2-3 bins with a mean and a sdev -- with three couplings: epsilon and zeta jointly pick the BI/BII bin, beta is binned on the previous residue's BI/BII, and chi plus the sugar torsions are binned on the sugar pucker. The following changes were made to support both RNA and DNA:
   * Three more couplings were added: alpha x gamma, BI/BII given pucker, and alpha(i+1) given BI/BII
   * R3 DNA potential only penalized deviation, not the bin; added a -log(P) bin penalty to the combined potential

The scoreterms: `na_torsion` and `na_torsion_well`.

Parameters were re-fit to non-redundant high-resolution (<= 2.3 A) RNA-only and protein/DNA datasets, regenerable via tmol/support/scoring/na_torsion_param_import.py.  txt files show the PDBs used in generating these statistics.

## na packing

The potential supplies a chi mean and sdev per bin, so rotamer generation falls out of it directly: every chi the sampler emits sits at a minimum of the term that scores it. Nucleic acids can adopt syn and anti; following Rosetta, RNA samples both and DNA only anti. 

## other minor fixes

- Fixed a bug where removal of terminal residue (water or ion) caused misannotation of termini
- Fixed a mismatch in numbering of jumps (root jumps should be UNnumbered)
- Set undernbuildable leaf atoms to NaN rather than 0,0,0
- torch.device("cuda") != torch.device("cuda:0"): added resolve_device() at the public
  entry points, mapping bare "cuda" to the current device
- patches gained `applies_to` / `add_torsions` / `add_chi_samples`


tmol/score/na_torsion/na_torsion_energy_term.py
```diff
@@ -0,0 +1,743 @@
...
+        return eval_na_torsion_for_pose
+
+    def get_rotamer_score_term_function(self):
+        return eval_na_torsion_for_rotamers
+
+    def get_score_term_attributes(self, pose_stack):
+        pbt = pose_stack.packed_block_types
+        p = self.params
+        bt = pose_stack.block_type_ind64
+        has_dna = bool(
```
kierandidi: this should be has_na right since rna also counts? or make sure we do not count rna at 4-7?

tmol/score/na_torsion/params.py
```diff
@@ -0,0 +1,206 @@
...
+
+# one index space for both polymers: DNA occupies 0-3 and RNA 4-7, so the
+# polymer of a base is base >> 2 and the tables that vary only by polymer are
+# gathered with that shift.
+POLYMERS = ("dna", "rna")
+POLYMER_IND = {name: i for i, name in enumerate(POLYMERS)}
+BASES = {"dna": ("a", "c", "g", "t"), "rna": ("a", "c", "g", "u")}
+N_BASE_PER_POLYMER = 4
+N_BASE = len(POLYMERS) * N_BASE_PER_POLYMER
+BASE_FOR_NAME3 = {
```
kierandidi, line 25: is it robust to match only based on 3letter name or should we also match on atoms? Just thinking since files might have A/C or so for dna as well or no?

tmol/score/na_torsion/na_torsion_energy_term.py
```diff
@@ -0,0 +1,743 @@
...
+        ang = tors[..., tor]
+        w = torch.stack([w_bi, 1.0 - w_bi], dim=-1)
+        e_bb = e_bb + torch.where(
+            both,
+            blended_devsq(ang, means[:, tor, :2], w) / sdev_bb[:, tor] ** 2,
+            zero,
+        )
+
+    # beta is binned on the previous residue's BI/BII state;
+    # fall back to BI mean when there is no preceding DNA residue
```
kierandidi: again valid for general NAs right?

tmol/score/na_torsion/potentials.py
```diff
@@ -0,0 +1,97 @@
...
+
+def triple_bin_weights(angle, means, sdev):
+    """Soft g+/t/g- assignment, replacing Rosetta's hard 120-degree bins."""
+    dev = wrap_degrees(angle.unsqueeze(-1) - means)
+    return torch.softmax(-dev * dev / (2.0 * sdev * sdev), -1)
+
+
+def bi_bii_weight(epsilon, zeta):
+    """Weight on BI. Rosetta's sigmoid on sin(epsilon - zeta), sharpness 20."""
+    return torch.sigmoid(-40.0 * torch.sin(wrap_degrees(epsilon - zeta) * RAD))
```
kierandidi, line 94: do we want to use sharpness 20 or -40 here? Code and docstring seem to be different?

tmol/tests/nucleic/test_na_scoring.py
```diff
@@ -0,0 +1,125 @@
...
+NA_ACTIVE_TERMS = (
+    ScoreType.fa_ljatr,
+    ScoreType.fa_ljrep,
+    ScoreType.fa_lk,
+    ScoreType.fa_elec,
+    ScoreType.hbond,
+    ScoreType.lk_ball,
+    ScoreType.lk_ball_iso,
+    ScoreType.cart_lengths,
+    ScoreType.cart_angles,
```
kierandidi, line 33: about about the new terms you introduced like na_torsion and na_torsion_well here? they also should be non-zero right?

tmol/pack/build_missing_sidechains.py
```diff
@@ -40,6 +46,8 @@ def build_missing_sidechains(
         pose_stack: The pose stack to process.
         sfxn: Score function used for packing.
         dunbrack_sampler: DunbrackChiSampler configured from the parameter DB.
+        na_sampler: NaChiRotamerSampler configured from the parameter DB; when
```
kierandidi, line 49: is this also working with pose_stack_from_pdb or only from biotite? 

Shall we deprecate the seperate from pdb and just make sure that this behidn the scenes calls the biotite pdb/cif parser?

tmol/pack/rotamer/na_chi_sampler.py
```diff
@@ -0,0 +1,320 @@
...
+            torch.arange(n_poses, device=poses.device).view(-1, 1, 1) * max_n_atoms
+            + index
+        )
+        xyz = poses.coords.reshape(-1, 3)[flat.reshape(-1)].reshape(*index.shape, 3)
+        xyz = torch.where(ok.unsqueeze(-1), xyz, torch.zeros_like(xyz))
+        w = pucker_weights(
+            xyz.reshape(-1, 5, 3), self.params.pucker_temperature
+        ).reshape(*bt.shape, N_PUCKER)
+        pucker = w.argmax(dim=-1)
+        return torch.where(ok.all(-1), pucker, torch.full_like(pucker, -1))
```
kierandidi, line 212: what happens if there is an incomplete sugar like in the pdb often? Is it just drops/does not get a rotamer?

tmol/tests/kinematics/test_fold_forest.py
kierandidi: shall we remove that print here?

kierandidi (approved):
Great, looking forward to have this in! runs for me and seems stable, just some nitpicks on code naming etc

tmol/score/na_torsion/na_torsion_energy_term.py
fdimaio: yeah ... this was originally DNA only and then converted to hybrid, not everything got renamed

tmol/score/na_torsion/params.py
fdimaio: it's tricky.  old structures use 'A'/'C'/'G'/'T' (or 'ADE' etc.) for both so only atom-based would get it correct.
but there are also legitimate RNA structures with unresolved O that would get resolved as DNA.
I lean toward using name as main decider?  But I'm open to other options

tmol/score/na_torsion/na_torsion_energy_term.py
fdimaio: yeah

tmol/score/na_torsion/potentials.py
fdimaio: will clarify the comment. both are right but unintuitive

Rosetta:
```
Real const sharpness = 20.0;
p1 = exp(-sharpness * sin(dev));
p2 = exp(+sharpness * sin(dev));
return p1 / (p1 + p2);
```

tmol/tests/nucleic/test_na_scoring.py
fdimaio: yes will update

tmol/pack/build_missing_sidechains.py
fdimaio: it works on any packer call.  so relax calls it everywhere.  biotite path calls it ONLY if atoms are missing.

fdimaio: i think discussion was to make 'pose_stack_from_pdb' a testing-only function

tmol/pack/rotamer/na_chi_sampler.py
fdimaio: residue is dropped at readin in this particular case.

we should discuss this, we can definitely support rebuilding more than we do.  but beyond the scope of this PR.

tmol/tests/kinematics/test_fold_forest.py
fdimaio: will do

PR #406, opened 2026-08-10
Better posestack minimizer

fdimaio:
Change LBFGS minimizer so that separate minimizer parameters are computed over each pose in the stack.  This avoids bad behavior when high- and low-energy poses are both minimized in a single stack.

Adds a field to the tmol minimizer, 'supports_segments' that when enabled look for an additional segment specifier used to split the minimization.  Each segment gets separate: a) hessian; b) convergence checks; and c) line search.

The search is still parallel (so no looping overhead) BUT the independent line searches and convergence checks mean more scorefunction evaluations are needed, adding a small but non-trivial overhead.

Note that other torch minimizers that do not have this field will still work but are behaved as before, minimizing sum of energies over the entire stack.

### DIFF
--- a/tmol/optimization/lbfgs_armijo.py
+++ b/tmol/optimization/lbfgs_armijo.py
@@ -1,5 +1,4 @@
 import torch
-from functools import reduce
 from types import SimpleNamespace
 from torch.optim import Optimizer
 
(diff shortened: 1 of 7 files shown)


aleaverfay (approved):
LGTM

PR #408, opened 2026-08-12
Add pose manipulation functions to tmol interface

fdimaio:
Adds 4 new functions:
- extended_pose_stack_from_sequences - generate a posestack _with extended backbone coordinates_ from seq.  Following R3, 1-letter-codes specify AAs/DNA, while the bracket format "A[ALA:nterm]" or "a[RA]" or "X[ATP]" allow specification of specific residue types by full name.  ':' separates chains while '(CCO)' allows smiles strings to automatically have params generated and exported.
- get|set_named_torsion - allows getting and setting torsions in a pose by name.  optionally take a foldtree, will use generate_simple_foldtree otherwise
- get_torsion_names - list all named torsions in a residue type

Some functions moved to new places to be more easily used in multiple contexts.

Renamed some functions in poseStackBuilder for consistency.


tmol/pose/build_context.py
```diff
@@ -0,0 +1,29 @@
+from dataclasses import dataclass, field
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
```
aleaverfay: What's going on here with the conditional import?

tmol/pose/pose_stack_builder.py
```diff
@@ -123,288 +132,102 @@ def i64(t):
...
-        n_atoms[real_res] = pbt.n_atoms[block_type_ind64[real_res]]
-        block_coord_offset = exclusive_cumsum2d(n_atoms)
-
-        max_n_atoms = torch.max(torch.sum(n_atoms, dim=1)).item()
-
-        chain_labels = numpy.full(block_type_ind64.shape, "", dtype=object)
-        real_res_np = real_res.cpu().numpy()
-        chain_labels[real_res_np] = "A"
+        if packed_block_types is None:
+            packed_block_types = cls._packed_block_types_for_names(
```
aleaverfay: ooo, this feels like it could lead to some inefficiencies -- each new PackedBlockType object has to be annotated before scoring can take place. I feel like the default should be to use the full residue type set to define the PackedBlockTypes object. If you make a pose from sequence of poly-valine and then run design, you'd be real surprised that it was unable to design to anything besides valine.

aleaverfay (approved):
I had some minor changes I thought would be worth making, but I'm gonna give the approval so you can merge once those are in 👍

fdimaio (commented):
Also, the test failure is probably my bad?  bounds should probably be bigger

fdimaio:
> Also, the test failure is probably my bad? bounds should probably be bigger

crap wrong page

tmol/pose/build_context.py
fdimaio: sorry, removed

tmol/pose/pose_stack_builder.py
fdimaio: good catch.... fixed (uses default db now)

PR #409, opened 2026-08-12
Aleaverfay/deprecate complex dispatch

aleaverfay:
Remove old dispatch versions; everything uses DeviceOperations now.


fdimaio (approved):
lgtm.  you might need to update CMakeLists with deleted files?

fdimaio:
> lgtm. you might need to update CMakeLists with deleted files?

Also I think the test failure is my fault, the bounds check should probably be bigger

It might not be the best test .... small numeric differences in step 1 can have pretty big effects by step 100....

aleaverfay:
> It might not be the best test .... small numeric differences in step 1 can have pretty big effects by step 100....

It is exactly the test that I would have written, but because of the GPU's non-deterministic operational ordering and the fact that summing a set of identical numbers in a different order gives different answers, it seems entirely reasonable that the trajectories could diverge, and then there's really no good way to reason about how much divergence is expected, it's very hard to choose an epsilon. If you look at the results that ran for this one, the stacked minimization improved the energy for Pose #2 by 9kcal/mol over the one-at-a-time minimization and there's no reason to believe that it couldn't have happened the other way around, with the one-at-a-time producing a 9kcal/mol difference.

fdimaio:
> > It might not be the best test .... small numeric differences in step 1 can have pretty big effects by step 100....
> 
> It is exactly the test that I would have written, but because of the GPU's non-deterministic operational ordering and the fact that summing a set of identical numbers in a different order gives different answers, it seems entirely reasonable that the trajectories could diverge, and then there's really no good way to reason about how much divergence is expected, it's very hard to choose an epsilon. If you look at the results that ran for this one, the stacked minimization improved the energy for Pose #2 by 9kcal/mol over the one-at-a-time minimization and there's no reason to believe that it couldn't have happened the other way around, with the one-at-a-time producing a 9kcal/mol difference.

When I first checked this in, max diff was ~2.  the code checks for abs(delta)<5.  Maybe mark as xfail and I can revisit this later?

PR #410, opened 2026-08-13
Add a method to dump a multi-model PDB string of a RotamerSet

aleaverfay:
Mainly just extracting code from a unit test and turning it into a usable class method

### DIFF
--- a/tmol/pack/rotamer/rotamer_set.py
+++ b/tmol/pack/rotamer/rotamer_set.py
@@ -1,4 +1,5 @@
 import attr
+import numpy
 import torch
 from tmol.types.attrs import ValidateAttrs
 from tmol.types.torch import Tensor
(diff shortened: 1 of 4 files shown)


tmol/pack/rotamer/rotamer_set.py
```diff
@@ -51,3 +52,39 @@ def _pose_ind_for_atom(self):
     @property
     def n_rotamers_total(self):
         return self.block_ind_for_rot.shape[0]
+
+    def write_pdb(self, pbt) -> str:
```
fdimaio: Looks good, but is this intended to be public facing or a private function for development?

If the former, it might make more sense to have this function write AtomArrays instead?  since we should really use biotite as our only public path for parsing.

If the latter, than '_' the name?

fdimaio (approved):
LGTM

tmol/pack/rotamer/rotamer_set.py
aleaverfay: Okay, I can give that a shot.

PR #411, opened 2026-08-13
Remove the code path that allowed PDBInfo to hold None arrays

aleaverfay:
Remove the code path that allowed PDBInfo to hold None for atom occupancies and bfactors.

The class itself does not hint that None is a valid option for these tensors, and it does not truly make sense to preserve the idea that there might not be values for these when having a default value is totally reasonable / workable.

If the user passes None in for these arguments in pose_stack_from_canonical_form, instead, a numpy array of the appropriate shape and filled with the default values will be created. Thereafter, anyone interacting with these arrays need not worry that they are actually None.

### DIFF
--- a/tmol/io/pose_stack_construction.py
+++ b/tmol/io/pose_stack_construction.py
@@ -5,7 +5,7 @@ from tmol.types.torch import Tensor
 from tmol.types.array import NDArray
 from typing import Optional
 from tmol.types.functional import validate_args
-from tmol.pose.pdb_info import PDBInfo
+from tmol.pose.pdb_info import PDBInfo, DEFAULT_ATOM_OCCUPANCY, DEFAULT_ATOM_B_FACTOR
 from tmol.pose.pose_stack import PoseStack
 from tmol.pose.packed_block_types import PackedBlockTypes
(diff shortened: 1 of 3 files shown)


tmol/io/pose_stack_construction.py
```diff
@@ -5,7 +5,7 @@
 from tmol.types.array import NDArray
 from typing import Optional
 from tmol.types.functional import validate_args
-from tmol.pose.pdb_info import PDBInfo
+from tmol.pose.pdb_info import PDBInfo, DEFAULT_ATOM_OCCUPANCY, DEFAULT_ATOM_B_FACTOR
```
fdimaio, line 8: Hi Andrew, very silly request but can DEFAULT_ATOM_B_FACTOR = 30.0
instead of 0?

B factors of 0 are badly behaved in crystal refinement/MR/etc., 30 is a very reasonable "average"

fdimaio (approved):
LGTM (see 1 comment)

tmol/io/pose_stack_construction.py
aleaverfay: Let's do it! I have no strong feelings.

PR #412, opened 2026-08-18
File naming refactor

aleaverfay:
I want to switch tmol's user interface pattern to one I see commonly in python packages: the important classes and functions are imported into the __init__.py files and the source files are named with a leading underscore. In order to make this happen, the first step is to clean up the inter-package dependencies. Claude has helped me untangle some tricky knots, and soon the file renaming will commence.


tmol/ligand/_fragmentation.py
```diff
@@ -843,6 +864,414 @@ def apply_fragment_connections(pose_stack, mapping: FragmentedLigandPoseMapping)
...
+        bt_idx = int(pose_stack.block_type_ind64[pose_ind, block_ind].item())
+        frag_bt = pbt.active_block_types[bt_idx]
+        orig_name = frag_bt.name.rsplit(".", 1)[0]
+
+        key = (pose_ind, orig_name)
+        if key not in group_ind_map:
+            pose_group_counter.setdefault(pose_ind, 0)
+            group_ind_map[key] = pose_group_counter[pose_ind]
+            pose_group_counter[pose_ind] += 1
+        group_ind = group_ind_map[key]
```
kierandidi: what happens here if we have two ligands with the same ligand name ie LG1 or so? Seems like in this setup here orig_name is a block type, but not a specific block instance; so if we had let's say the same ligand type twice in a pose it will be merged into one restored block, and if the atoms have the same name we would overwrite coords etc? Probably we should carry an explicity group ID/source residue ID or so?

tmol/pose/_pose_stack_builder.py
kierandidi: what about the framgentation/spit_block_mapping metadata here? We need to carry that through if present so that we could restore/merge batched fragmented poses later on again right?

tmol/ligand/_fragmentation.py
```diff
@@ -843,6 +864,414 @@ def apply_fragment_connections(pose_stack, mapping: FragmentedLigandPoseMapping)
...
+                e = entry_lookup[(p, old_b)]
+                n_conn = len(pbt.active_block_types[e.orig_block_type_ind].connections)
+            else:
+                n_conn = len(
+                    pbt.active_block_types[
+                        int(pose_stack.block_type_ind64[p, old_b].item())
+                    ].connections
+                )
+            for c in range(n_conn):
+                partner = old_irc[p, old_b, c]
```
kierandidi: here we do the merging IIUC, but we only map the first fragment to its restored block right? the index c here is not always the overall connection index to the parent or is it?

tmol/ligand/_fragmentation.py
```diff
@@ -843,6 +864,414 @@ def apply_fragment_connections(pose_stack, mapping: FragmentedLigandPoseMapping)
...
+    old_pdb = pose_stack.pdb_info
+    old_bco = pose_stack.block_coord_offset.cpu().numpy()
+    chain_np = np.full((n_poses, new_max_n_blocks), -1, dtype=np.int32)
+    res_labels = np.zeros((n_poses, new_max_n_blocks), dtype=int)
+    ins_codes = np.full((n_poses, new_max_n_blocks), "", dtype=object)
+    chain_labels = np.full((n_poses, new_max_n_blocks), "", dtype=object)
+    occ = np.full((n_poses, new_max_n_atoms), DEFAULT_ATOM_OCCUPANCY, dtype=np.float32)
+    bf = np.full((n_poses, new_max_n_atoms), DEFAULT_ATOM_B_FACTOR, dtype=np.float32)
+    old_chain = pose_stack.chain_id.cpu().numpy()
+    for p, (blocks, m) in enumerate(zip(per_pose_blocks, old_to_new)):
```
kierandidi: similar thing here if I am not mistaken? the group of fragments has same metdata called orig_*, and we should use that to populate things like residue, chain etc right? Since right now this is not necessarily given for all fragments

tmol/ligand/_fragmentation.py
```diff
@@ -843,6 +864,414 @@ def apply_fragment_connections(pose_stack, mapping: FragmentedLigandPoseMapping)
...
+        for old_b, new_b in m.items():
+            if new_b < 0:
+                continue
+            chain_np[p, new_b] = old_chain[p, old_b]
+            res_labels[p, new_b] = old_pdb.residue_labels[p, old_b]
+            ins_codes[p, new_b] = old_pdb.residue_insertion_codes[p, old_b]
+            chain_labels[p, new_b] = old_pdb.chain_labels[p, old_b]
+        for new_b, (bt_idx, kind, src) in enumerate(blocks):
```
kierandidi: what about things here that are not residue/chain labels like bfactor or occupancy or so?

tmol/ligand/_fragmentation.py
```diff
@@ -843,6 +864,414 @@ def apply_fragment_connections(pose_stack, mapping: FragmentedLigandPoseMapping)
...
+        inter_residue_connections=new_irc64.to(torch.int32),
+        inter_residue_connections64=new_irc64,
+        inter_block_bondsep=new_ibs64.to(torch.int32),
+        inter_block_bondsep64=new_ibs64,
+        block_type_ind=new_bt32,
+        block_type_ind64=new_bt64,
+        chain_id=new_chain_id,
+        chain_id64=new_chain_id.to(torch.int64),
+        pdb_info=new_pdb_info,
+        split_block_mapping=None,
```
kierandidi: again there is something missing I think when remapping, we need the remap the atom indices or the constraint block at this point here no?

tmol/score/_score_utils.py
```diff
@@ -1,15 +1,6 @@
 import torch
```
kierandidi: why do we have this file once in tmol/score and once in tmol/ops? seems to be the same one, can we unify that?

tmol/ligand/__init__.py
```diff
@@ -1,54 +1,163 @@
...
+    _correct_ring_nitrogen,
+    _get_hyb,
+    _has_sp2_oxygen_neighbor,
+    _modify_polar_c,
+    _neighbor_counts,
+    assign_tmol_atom_types,
+    kekulize_tolerant,
+    sanitize_tolerant,
+    should_kekulize_for_typing,
+)  # noqa: F401
```
fdimaio, line 32: should we be exporting all of these functions in init?

fdimaio (approved):
This looks good.  but is the intent that everything exported through `__init__` is the public interface (and calling between layers is only through public functions?)

The reason I ask is that some of what is exported are not functions that I think we want to expose.  this is very apparent in tmol.ligand but in a few other places as well.

tmol/ligand/__init__.py
aleaverfay: The first pass implementation of this PR is blocking, and so our goal is to be somewhat rough in the first pass and refine in the second pass. Kieran has already made strides towards firming up our API through the use of the __all__ variable in the __init__.py files. I think we want to be explicit with "these are the functions we want users to access" and then when we have that list, we can go back through everything that is being exported but not used anywhere else in tmol and then remove those functions from the export statements.

PR #413, opened 2026-08-21
Build Sphinx docs and interactive TMol tutorials

kierandidi:
## Stack

- depends on #415 (`kdidi/tmol-api-cleanup`)
- #415 is based on #412 (`aleaverfay/file_naming_refactor`)
- this PR contains the documentation/site layer and its tutorial data
- merge order: #412 → #415 → #413

## Summary

- replace the legacy generated API pages with a task-oriented Sphinx site for the public package layout established by #415
- document all 20 packages with explicit public APIs: 336 exported names are present in the rendered reference
- add eight executable tutorials covering structures, GPU batching, scoring, packing, minimization, FastRelax, ligands, and nucleic acids
- preserve live molecular viewers, selectable structures, tables, plots, exercises, notebook downloads, and Colab entry points
- use a restrained five-item header—Quickstart, Workflows, Examples, API, and Contributing—with the notebook gallery directly under Examples
- use static captures of the tutorials' rendered proteins, ligands, and nucleic acids as all eight gallery thumbnails
- follow the RF4 notebook pattern: collapse setup and presentation-only inputs while keeping modeling code and every output visible
- align objectives, prerequisites, observations, exercises, and explicit parity caveats with official Rosetta/PyRosetta tutorials
- clarify score-unit boundaries, constraint weighting, FastRelax behavior, ddG limitations, ligand parameter round trips, and scientific smoke-test versus convergence claims
- remove obsolete docs machinery, unreferenced assets, and unused/redundant docs dependencies
- build previews for stacked PRs in addition to PRs targeting the main branch
- allow 180 seconds for Jupyter kernel startup on busy GPU runners while retaining the 600-second per-cell execution timeout

## Test plan

- [x] strict Sphinx HTML build (`-W --keep-going -E`)
- [x] strict Sphinx link check
- [x] execute all eight notebooks on CPU-compatible paths
- [x] GPU-only batching and FastRelax cells remain covered by Docs CI
- [x] public API audit: 20 packages / 336 exports / 0 missing
- [x] public API regression tests (22 passed)
- [x] Colab release/configuration tests (4 passed)
- [x] notebook JSON, empty-source-output integrity, and extracted-code Ruff checks
- [x] pre-commit formatting and lint checks across the repository
- [x] rendered-browser audit of navigation, all eight thumbnails, API rows, and every code toggle
- [x] Docs GPU execution and preview publish for the latest stacked head
- [x] full CI

Preview: https://uw-ipd.github.io/tmol/previews/pr-413/

PR #414, opened 2026-08-22
Support common phosphorylated and methylated residues

kierandidi:
Stacked on #413. This is the first PTM slice and remains below 1k added lines (463 additions, 31 deletions).

## What changes

- Adds Rosetta-compatible SEP, TPO, PTR, MLZ, MLY, and M3L residue variants.
- Lets a patched residue expose an external name3 while retaining its canonical parent I/O equivalence class.
- Preserves PTM names through Biotite round trips and accepts OP1/OP2/OP3 phosphate aliases.
- Adds Rosetta-derived ideal geometry and partial charges with finite beta2016 gradients.
- Recognizes built-in PTMs before mixed ligand preparation, so prepare_ligands=True does not register SEP again.
- Updates canonical-ordering invariants for multiple external names sharing one I/O class.

## Validation

- 13 PTM tests pass on CPU; CUDA coverage is exercised by the top workflow PR.
- The full stacked I/O suite passes: 178 passed, 138 CUDA variants skipped on the CPU host.
- PyRosetta 2026.33 generated all six native PTM variants; ideal new-bond geometry agrees to below 0.000002 Angstrom before PDB rounding.
- Ruff formatting/check and git diff checks pass.

## Stack

Merge #412, then #413, then this PR.


kierandidi:
Rebased this foundation onto the current `kdidi/sphinx-docs-refactor` head (`13e06ed31`); the PTM commit is now `bb0abc515` and GitHub reports it mergeable. `git range-diff` shows the PTM patch itself is unchanged. Downstream post-rebase H200 validation passes the 34-case PTM parity/workflow suite and the 17-case explicit-metal stack; the glycan stack is in its final dependent rerun.

PR #415, opened 2026-08-23
Clean up TMol public APIs and shared scoring helpers

kierandidi:
## Stack

- based on #412 (`aleaverfay/file_naming_refactor`)
- followed by #413, which adds the Sphinx site and tutorials on top of this branch
- merge order: #412 → #415 → #413

## Summary

- define intentional public imports across the top-level TMol packages and cover them with public-API regression tests
- make `calculate_fragment_interactions` an explicit public `tmol.score` export for the scoring-analysis workflow
- register the backbone-torsion score-term creator explicitly instead of discovering score-term modules through a filesystem glob
- keep fragment-interaction scoring in a focused score module while consolidating shared scoring/minimization helpers under `tmol.ops`
- remove the duplicated 856-line `tmol.score._score_utils` implementation, its duplicate test file, an unused chain-mask helper, and stale debug/PDB-writing code
- make coordinate-constraint construction handle non-protein blocks and use the correct parameter shape without debug output
- add optional notes to `selection_gallery()` and test the rendered selection metadata
- update affected ligand, ddG, constraint, score-term, visualization, and operations tests
- correct coverage exclusions for the file-naming refactor

## Test plan

- [x] `pre-commit run --all-files`
- [x] public API regression tests (22 passed)
- [x] focused visualization, constraint, score-term, operations, ligand, and ddG tests
- [x] full CI


tmol/ops/_score_utils.py
```diff
@@ -7,27 +7,6 @@
 from tmol.optimization import run_cart_min
 
 
-def residue_mask_from_chain(pose_stack, chain_id):
```
aleaverfay, line 10: Why delete this function?

aleaverfay (approved):
This all makes sense to me, except the removal of the residue_mask_from_chain utility. It probably belongs as something like a residue selector, but could rather harmlessly persist in this file for now.

fdimaio (approved):
lgtm

PR #436, opened 2026-08-27
Fix heterogeneous dense minimizer segments

kierandidi:
Treat contiguous segments as dense only when every segment has the same length. Unequal mutation legs otherwise take the reshape fast path and fail during batched kinematic minimization. Includes a focused regression test and bumps the patch version to 0.1.50.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,7 +21,7 @@ classifiers = [
     "Programming Language :: Python :: Implementation :: CPython",
     "Topic :: Scientific/Engineering :: Bio-Informatics",
 ]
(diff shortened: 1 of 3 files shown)


kierandidi:
Validation:

- Focused CPU regression passes for contiguous unequal segment sizes (4 and 3 DOFs).
- Existing segmented-vs-independent minimization regression passes.
- Reproduced the downstream failure before the fix with segment counts `[74, 72]` (`reshape [2, 74]` from 146 values).
- The same real two-pose protein-ligand mutation relaxation completes after the fix while remaining batched.

PR #438, opened 2026-08-28
docs: document macOS build paths

ssiddhantsharma:
## Summary

The AOT-vs-JIT documentation and the `extensions are not built` error message state that JIT mode **requires `nvcc`**. That is true only for the **CUDA** kernels — `cuda_if_available()` drops the `.cu` sources when CUDA is unavailable, so CPU-only platforms (e.g. macOS / Apple Silicon) compile the kernels with just a C++ compiler and `ninja`, no `nvcc`. The current wording steers Mac/CPU users away from the one path that works for them.

Verified on macOS-ARM (Apple Silicon, CPU-only, no CUDA toolkit):
- `TMOL_USE_JIT=1 python -c "import tmol"` compiles the CPU kernels via clang and imports cleanly — no `nvcc`.
- `pip install -e . -Ccmake.define.TMOL_ENABLE_CUDA=OFF` (CMake in `>=3.18,<4`) builds loadable AOT CPU kernels; `tmol._cpp_lib._ensure_loaded()` succeeds without JIT.
- A plain `pip install -e .` on a machine with no CUDA toolkit can install without loadable CPU kernels, which is the trap this clarifies.

## Changes (docs + one error string)

- **DEVELOPMENT.md** — correct the JIT bullet and the *CUDA toolkit for JIT mode* section (nvcc is only for CUDA kernels; CPU-only JIT needs a C++ compiler + `ninja`); add a macOS / CPU-only row to the scenarios table.
- **docs/installation.md** — note that `TMOL_ENABLE_CUDA=OFF` is required when no CUDA toolkit is present, add a `TMOL_USE_JIT=1` alternative, and fix the "if nvcc is available" comment.
- **tmol/_cpp_lib.py** — correct the user-facing `TmolExtensionNotBuiltError` / loader-error guidance so CPU-only users are pointed at the working paths.

Docs/comment-only; no behavior change.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -56,7 +56,7 @@ tmol's C++/CUDA kernels can be loaded in two ways:
 
 - **AOT (Ahead-Of-Time)**: Pre-compiled `.so` libraries are bundled inside the installed package (e.g., from a wheel). Operations are registered in `torch.ops.tmol_*` namespaces. This is the default and requires no compiler at runtime.
 
-- **JIT (Just-In-Time)**: Source files (`.cpp`, `.cu`) are compiled on first use via `torch.utils.cpp_extension.load()`. This requires `nvcc` and a C++ compiler to be available. Useful for kernel development where you want to edit and reload C++/CUDA code without rebuilding the whole package.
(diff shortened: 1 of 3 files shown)


kierandidi:
Audited against current master and #465. The useful behavior here is now covered more completely by #465: automatic CPU-only fallback when nvcc is absent, macOS arm64 CPU wheels, native-vs-release CUDA architecture semantics, current PyTorch/C++ standard requirements, and PyTorch 2.14 coverage. This branch is also conflicting and has no current CI. I recommend closing this as superseded once #465 lands rather than rebasing a second copy of the same build documentation.

PR #442, opened 2026-08-28
Clarify TMol Python API contracts for v0.1.52

kierandidi:
## Summary

This is a deliberately behavior-preserving cleanup of TMol's Python surface, kept separate from the performance work in #441.

- adds concise Google-style documentation to important public APIs, database schemas, and previously empty product package boundaries
- applies TMol's native shape-aware `Tensor[dtype][shape]` and `NDArray[dtype][shape]` annotations to high-value I/O, pose, kinematics, minimization, rotamer, and coordinate-building interfaces
- removes stale commented implementations, obsolete pre-Python-3.9 compatibility code, redundant temporaries, and low-risk control-flow noise
- tightens runtime shape/type utilities, including fixing integer dimension validation to compare by value rather than object identity and adding a regression for large dimensions
- documents tensor layouts where batch/block/atom dimensions are not obvious

No new typing dependency is introduced, no scoring kernel behavior is changed, and dynamic/private implementation code was not mass-annotated merely to satisfy a metric. The diff spans 105 files because 51 package initializers receive one-line module documentation; overall it is +655/-631 (net +24 lines).

## Validation

- repository-wide Black check: 481 non-vendored Python files clean
- repository-wide Flake8 check: 481 non-vendored Python files clean
- Ruff and `git diff --check`: clean
- exact-revision AOT wheel built successfully (CPython 3.12, CUDA/H200)
- 67 exact-AOT I/O and minimization tests passed on CPU/GPU
- focused subsystem runs passed for chemical schemas, pose stacks, move maps, optimizer internals, type utilities, conformer samplers, coordinate construction, and the low-risk simplifications
- 55 product packages imported successfully and expose non-empty module documentation

The cleanup branch is based directly on current `master`; #441 has only one overlapping file, with independent changes on separate lines.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,7 +21,7 @@ classifiers = [
     "Programming Language :: Python :: Implementation :: CPython",
     "Topic :: Scientific/Engineering :: Bio-Informatics",
 ]
-version = "0.1.51"
+version = "0.1.52"
 
 dependencies = [
     # Core ML framework
--- a/tmol/__init__.py
+++ b/tmol/__init__.py
@@ -1,3 +1,5 @@
+"""TMol: GPU-accelerated molecular modeling with PyTorch."""
+
 # flake8: noqa
 
 # Load pre-compiled C++/CUDA extensions (TORCH_LIBRARY ops).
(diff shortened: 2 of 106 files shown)


kierandidi:
Release preparation: this PR now includes the single pyproject.toml version bump from 0.1.51 to 0.1.52, matching the established TMol tag-triggered release process. The refreshed CI will therefore validate the exact tree that will be tagged.

PR #473, opened 2026-09-08
Reduce idle LBFGS mask work

kierandidi:
## Summary

- reuse the existing per-segment LBFGS reset-mask tensors instead of allocating replacement masks on every ordinary iteration
- track whether the previous reset mask actually needs clearing, preserving the existing conservative migration behavior for older optimizer state dictionaries
- skip constructing the converged/stalled union when the existing host mirror proves that no segment is inactive

The change does not alter the optimizer API, line-search policy, score functions, tensor layouts, backend dispatch, or numerical operations on active/reset segments.

## Performance

For 12-iteration Cartesian minimization profiles on H200 (5uoi and hsp90, batch 1/4, eager/CUDA Graph):

- 24 fewer physical CUDA launches at every one of the 8 measured points versus current master
- 12 fewer `aten::empty_strided`, 12 fewer `aten::zero_`, and 12 fewer inactive-mask launches
- unchanged peak GPU memory (apart from an immaterial 512-byte profiler-level difference in the first step)
- the incremental inactive-mask guard measured 1.008x geomean; direct paired A/B timing is running in parallel because launch counts are the stable acceptance signal for this small change

CPU reversed-process A/B across 43/548 residues and 1/8 threads was neutral: 0.999x point geomean.

## Validation

- focused optimizer CPU suite: 30 passed, 21 skipped, 1 xfailed
- exact single-thread score and coordinate checksums; multithread deltas remain within existing reduction-order noise
- dedicated storage-reuse regression test added

Benchmark and profiler harnesses/results remain outside the repository.

### DIFF
--- a/tmol/optimization/_lbfgs_armijo.py
+++ b/tmol/optimization/_lbfgs_armijo.py
@@ -362,6 +362,7 @@ class LBFGS_Armijo(Optimizer):
             if state.get(name) is not None:
                 state[name].zero_()
         state["any_needs_reset"] = False
+        state["any_was_reset"] = False
         state["any_inactive"] = False
         self._last_loss_vec = None
         self._closure_fn = None
@@ -581,6 +582,7 @@ class LBFGS_Armijo(Optimizer):
             state["stalled"] = torch.zeros(self._n_segments, **flags)
             state["needs_reset"] = torch.zeros(self._n_segments, **flags)
             state["was_reset"] = torch.zeros(self._n_segments, **flags)
+            state["any_was_reset"] = False
 
         return SimpleNamespace(
             # config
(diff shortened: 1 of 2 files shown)


kierandidi:
Final combined post-merge A/B completed successfully. Across 5uoi/hsp90, B1/B4, and eager/CUDA Graph, the geometric mean of the 8 point-median speedups is **1.0123x**. Seven point medians improve (up to 1.0285x); hsp90 B1 graph is neutral at 0.9997x. Exact profiles remove 24 physical CUDA launches at every point versus #472, and peak allocation is effectively unchanged. The reset-mask half alone was neutral (0.9979x CUDA, 0.9988x CPU); the inactive-mask guard provides the repeatable combined benefit. Raw benchmark and profile artifacts remain outside the repository.

PR #474, opened 2026-09-08
Skip excluded CUDA LJ/LK atom pairs

kierandidi:
## Summary

- return before coordinate, distance, and parameter work for CUDA LJ/LK atom pairs excluded by count-pair separation
- cover fused energy-only, weighted derivative, and full energy/derivative helpers used by pose and rotamer scoring
- specialize at compile time so CPU keeps the established path; no score-term, weight, layout, or public-API policy changes

The default fused LJ/LK-electrostatics traversal already has the equivalent joint shortcut. This change targets independent LJ/LK scoring used by term toggles, block-pair outputs, rotamer scoring, and general fallback paths.

## Performance

A three-repeat reversed-order H200 matrix over 5uoi and 5yzf at B1/B4 found all 24 paired LJ/LK measurements faster. Point medians improve by 10.47-12.37% for score-only and 2.61-3.15% for score plus gradient, with unchanged measured peak allocation. Exact score checksums match; CUDA gradient totals stay inside the baseline atomic-reduction repeat envelope.

The broad portable experiment was rejected: CPU LJ/LK regressed (0.983x geometric mean) and electrostatics was neutral on CUDA. This PR retains only the measured CUDA LJ/LK specialization. Benchmark harnesses and raw artifacts remain outside the repository.

## Validation

- corrected source compiled for CPU and CUDA against current master
- 62 focused LJ/LK tests passed in the broad semantic gate
- current-master repeated A/B, full focused tests, and hosted CI are running in parallel; this PR will not be merged until all finish cleanly

### DIFF
--- a/tmol/score/ljlk/potentials/ljlk.hh
+++ b/tmol/score/ljlk/potentials/ljlk.hh
@@ -530,12 +530,16 @@ TMOL_DEVICE_FUNC std::array<Real, 2> lj_atom_energy(
 
 // Fused LJ/LK energy for pose scoring. Heavy-atom pairs reuse the coordinate
 // loads, distance, and count-pair separation for both potentials.
-template <typename Real>
+template <typename Real, tmol::Device D>
 TMOL_DEVICE_FUNC std::array<Real, 3> ljlk_atom_energy(
     int atom_tile_ind1,
     int atom_tile_ind2,
     LJLKScoringData<Real> const& score_dat,
     int cp_separation) {
+  if constexpr (D == Device::CUDA) {
+    if (cp_separation < 4) return {0.0, 0.0, 0.0};
+  }
+
   using Real3 = Eigen::Matrix<Real, 3, 1>;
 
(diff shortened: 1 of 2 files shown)


kierandidi:
Clean current-master confirmation, first complete round (job 200714):

- CPU LJ/LK geometric mean: 1.003x (compile-time neutral path)
- H200 LJ/LK geometric mean across 5uoi/5yzf, B1/B4, score/gradient: 1.069x
- H200 score-only: +10.27% to +12.47%
- H200 score+gradient: +1.90% to +2.95%
- electrostatics control geometric mean: 0.997x (unchanged source)
- exact energy checksums and unchanged measured peak allocation; absolute-gradient checksums agree apart from the established CUDA atomic accumulation envelope

The reversed-order repeats continue, but this already reproduces the earlier 3-repeat broad experiment from the exact final commit. Hosted CPU, docs, lint, and Codecov are green; hosted CUDA is still running.

PR #475, opened 2026-09-08
Skip distant LJ/LK derivative work

kierandidi:
## Summary

- reject LJ/LK derivative atom pairs at the existing maximum-distance boundary before normalization and spline work
- reuse the atom-1 radial derivative with the opposite sign for atom 2
- preserve CPU and CUDA behavior, layouts, score-term toggles, live weights, and autograd interfaces

## Performance

Clean direct source A/B on the first 43-residue rotamer workload:

- one-thread CPU independent LJ/LK gradient: 5.184 s -> 3.297 s (1.57x)
- H200 B1 independent LJ/LK gradient: 7.856 ms -> 5.981 ms (1.31x)
- H200 B4 independent LJ/LK gradient: measured in the attached benchmark matrix
- forward scoring and peak memory are unchanged controls

A second 548-residue CPU/H200 point and hosted validation are running. Benchmark harnesses and raw artifacts remain outside the repository.

## Correctness

The early return matches the existing V_dV cutoff exactly (`dist >= max_dis`), and the atom-2 distance derivative is exactly the negative of atom 1. CPU score and gradient checksums are exact in the completed A/B point; CUDA differences remain within normal atomic accumulation nondeterminism.

PR #476, opened 2026-09-08
Fuse LJ/LK derivative traversal

kierandidi:
## Summary

- evaluate weighted LJ and LK derivatives in one all-atom pair traversal
- reuse coordinate loads, distance, count-pair separation, and coordinate atomics for heavy-atom pairs
- remove the duplicated LK-only traversal and all superseded zero-caller LJ/LK helper templates
- preserve the three independent upstream score-lane weights, public APIs, layouts, and fallback behavior
- make the CUDA shared-versus-independent rotamer-gradient check robust to the expected floating-point reassociation with an eight-epsilon scale-aware floor

## Performance

A clean two-order A/B against master after #475 produced all 24 expected records. Across 5uoi and 5yzf, one-thread CPU LJ/LK gradients improve by 1.121--1.129x (1.125x geometric mean); H200 B1/B4 gradients improve by 1.058--1.076x (1.066x geometric mean). Forward scoring is neutral at 1.00003x geometric mean and peak CUDA allocation is unchanged at every point. Entry counts and score checksums are exact; the tiny gradient checksum differences are expected from reassociation.

Combined with #475, the improvement over the #474 baseline remains about 1.76--1.81x on CPU and about 1.42x on H200.

## Validation

- focused CPU/CUDA suite: 77 passed, 3 skipped
- final A/B: 24/24 records with invariant entry counts and peak allocation
- the production diff removes 358 net lines across two implementation files and adds no backend, shape, or score-policy branch
- one stochastic CUDA atomic-order assertion failed on L40 and reproduced on H200; diagnostics measured a 0.0195 maximum difference against a 24,589 gradient scale (7.9e-7 relative). The test now retains its repeat envelope plus an eight-machine-epsilon scale-aware floor. Ten targeted H200 repetitions pass; the complete hosted matrix now gates merge.


kierandidi:
Focused validation completed before this update: **77 passed, 3 skipped** across CPU and CUDA LJ/LK tests plus score-function/block-pair/packing coverage. The follow-up cleanup removes four superseded, zero-caller LJ-only/LK-only helper templates, bringing the complete production diff to **59 additions and 417 deletions** across the same two files. It changes no instantiated path; the branch remains API-, lane-, weight-, layout-, and backend-policy-neutral.

PR #510, opened 2026-09-13
[2/10] Score general chemistry with explicit torsion ownership and connection parameters

kierandidi:
Extend scoring to general chemistry while preserving Frank’s force-field choices.

- **Correctness:** Partition torsion ownership, enumerate connection-centred impropers, preserve injected parameters and constraint precision, and fix junction references and correlated-group gradients. Unify parameter-annotation caching behind one bounded weak-identity cache used by every term, correct water to beta_nov16 radii, well depths and charges, and name the mirrored disulfide parameters instead of indexing them.
- **Efficiency:** Retain master’s fused scoring and sparse dispatch; compile junction aliases once, reuse tables and bound caches.
- **Validation:** CPU/CUDA scoring and derivative checks pass. GPU CI now detects faulty allocations and retries confirmed hardware failures; ordinary test failures remain failures.

Stack 2/10, based on #509; followed by #511. Derived from Frank DiMaio’s #503 and the fixes in #508, retaining master’s #507 efficiency work.


kierandidi:

PR #511, opened 2026-09-13
[3/10] Share ligand conversion and protonation with AtomWorks

kierandidi:
Replace duplicated molecular conversion and protonation with AtomWorks.

- **Correctness:** Preserve source atoms, charges, stereochemistry and explicit histidine policy. Share MOL2 preparation and reject undefined bond orders instead of guessing chemistry. Restore generic oxyacid inference and rebuilds for non-authoritative charge models, keep Dimorphite vendored, and reject residues with no resolved geometry.
- **Efficiency:** Remove duplicate protonation code; reuse compiled rules, scoped templates and bulk coordinate conversion.
- **Validation:** Molecular conversion/protonation workflows pass. AtomWorks parity covers 774 original states; stack CI continues.

Stack 3/10, based on #510; followed by #512. Derived from Frank DiMaio’s #503 and the fixes in #508, retaining master’s #507 efficiency work.

Requires AtomWorks [#349](https://github.com/baker-laboratory/atomworks-dev/pull/349) (shared conversion/protonation) and [#351](https://github.com/baker-laboratory/atomworks-dev/pull/351) (amide/phosphoryl completion and optional CCD reading). Pinned to `6b30ea05`; merge/release remains a landing dependency.

PR #512, opened 2026-09-13
[4/10] Build noncanonical polymers from chemistry-derived capped profiles

kierandidi:
Prepare noncanonical polymers from declared chemistry and connections.

- **Correctness:** Retain unresolved atoms and source identity, cap backbones for parameter generation, and reconstruct polymer types with terminal patches and nucleic-acid references.
- **Efficiency:** Cache immutable profiles, avoid unnecessary cap coordinates and assemble completed arrays once.
- **Validation:** Profile/cache checks pass; terminal GTP retains its three phosphates. Extended modified-polymer workflows are isolated in #518. Updated stack CI and the expanded CIF/FastRelax sweep are in progress.

Stack 4/10, based on #511; followed by #513. Derived from Frank DiMaio’s #503 and the fixes in #508, retaining master’s #507 efficiency work.

Requires AtomWorks [#349](https://github.com/baker-laboratory/atomworks-dev/pull/349) (shared conversion/protonation) and [#351](https://github.com/baker-laboratory/atomworks-dev/pull/351) (amide/phosphoryl completion and optional CCD reading). Pinned to `6b30ea05`; merge/release remains a landing dependency.


tmol/io/_atomworks_reader.py
```diff
@@ -0,0 +1,125 @@
...
+
+def read_cif(
+    path,
+    *,
+    model=1,
+    assembly_id=None,
+    author_fields=False,
+    extra_fields=None,
+    hydrogen_policy="rebuild",
+):
```
fdimaio: Is it intended that all poses go through 'read_cif' in atomworks parser?

reason I ask is there are two 'modes' in reading tmol:
1) components of the structure are not known at load-time (e.g., pose contains ligands tmol doesn't know about).  the parameter-gen pipe needs to run
2) all components of the structure are known at load-time

in 1, much of the info that atomworks provides IS needed -- protonation, bond typing, etc.

but in 2, all that is needed is res+atom names and coordinates

so the old code was "permissive" ... if (say) bond records were absent, or input was PDB format, it was fine as long as tmol knew all the parameters

I'd like to keep that if possible....
- read_cif() reads whatever info is provided. 
- use_ccd flag allows the model to optionally use CCD to fill in missing information
then atomarray->pose_stack checks what is needed.  if data is needed (say, new ligands bond-types) that isn't present, we throw an error saying that residue type XYZ needs add'l info.

One can provide this through a 'properly typed cif input'.  OR they can provide a ligand mol2/tmol and continue reading their PDB stack

tmol/io/_cif.py
```diff
@@ -0,0 +1,549 @@
...
+        templates needed for unresolved stereochemistry. Slicing, copying and
+        AtomWorks concatenation preserve these templates. Coordinate-only reading
+        returns a plain Biotite AtomArray.
+    """
+    if assembly_id is not None:
+        if not isinstance(assembly_id, str) or not assembly_id:
+            raise ValueError("assembly_id must be a nonempty string or None")
+        if not include_bonds:
+            raise ValueError("Assembly reading requires include_bonds=True")
+    if reader == "atomworks":
```
fdimaio: I'd rather not have this explicit if possible?  

reader takes any input and tries its best to return as much info as possible
- model/assmbly-id/use_ccd are options
- include_bonds/extra_fields/reader/Hpolicy are not
if tmol knows about a residue already all this extra info is unneeded

tmol/io/_atomworks_reader.py
kierandidi: Addressed in the updated stack (reader commit 97ee2ea4a; construction changes in #516). File reading now exposes only model, assembly_id and use_ccd. With use_ccd=False, it reads supplied atoms/bonds without dictionary lookup; known components use names and coordinates with an existing parameter database or ligand_params_files, without requiring generation. Unknown components are validated at preparation and the error names the residue and missing bond-order information. PDB CONECT alone is not treated as reliable bond orders for an unknown component. Supplied hydrogens are retained. CPU/CUDA integrated PDB/CIF and parameter-file workflows pass; all 44 reader/mirror checks and two CCD-disabled assembly checks pass. A full updated suite is running.

tmol/io/_cif.py
kierandidi: Simplified as requested: the public reader options are model, assembly_id and use_ccd. Removed reader/include_bonds/extra_fields/hydrogen_policy from the file API. The optional CCD policy is shared with AtomWorks; supplied chemistry is preserved, and preparation validates only what is needed for components without parameters. Published in the updated stack and covered by integrated CPU/CUDA tests.

kierandidi:

PR #514, opened 2026-09-13
[6/10] Integrate general residue and conjugate preparation with instance identity

kierandidi:
Integrate generalized ligand, conjugate and polymer preparation.

- **Correctness:** Preserve repeated-residue identity, fragments, authored chemistry and supplied attachment charges on reload. Restore terminal-cap forces and retain generated K300/K80 junction parameters for non-amine backbones exposed by 3SVU; unify MOL2 and single-ligand CIF entry paths. Add Frank’s MOL2 `keep`/`auto`/`regenerate` modes, and prepare/export canonical sidechain-to-sidechain links. Share reproducible CIF/MOL2 seeds and remove orphan hydrogens only during regeneration (9EWF).
- **Efficiency:** Reuse already-prepared chemistry and remove redundant cap-row copying while retaining master’s batched conformer minimization.
- **Validation:** Ligand/scoring, cap-force and neutralized phosphate/EOP workflows pass on CPU/CUDA, including minimization and parameter roundtrips. Difficult source fixtures are in #518. Updated stack CI and the expanded CIF/FastRelax sweep are in progress.

Stack 6/10, based on #513; followed by #515. Derived from Frank DiMaio’s #503 and the fixes in #508, retaining master’s #507 efficiency work.

Requires AtomWorks [#349](https://github.com/baker-laboratory/atomworks-dev/pull/349) (shared conversion/protonation) and [#351](https://github.com/baker-laboratory/atomworks-dev/pull/351) (amide/phosphoryl completion and optional CCD reading). Pinned to `6b30ea05`; merge/release remains a landing dependency.

PR #515, opened 2026-09-13
[7/10] Sample covalent groups coherently with bounded packing and chemical kinematics

kierandidi:
Pack and move covalently connected groups coherently.

- **Correctness:** Use joint conformers, pose-scoped identity and chemistry-derived trees; enforce explicit sample budgets before allocation. Default FastRelax uses its scoring database, nucleic-acid rotamers and joint attached-group sampling; frozen chi preserve input angles. Declare side-chain sampling as one bundle, ConjugatedSCSampler(SidechainSampler.from_database(...)), and trigger the fallback sampler on the rotamers other samplers actually built rather than on what they declare.
- **Efficiency:** Bound explosive supplemental sampling while retaining canonical proton grids; avoid duplicate traversal, independent fallback rows and per-rotamer CUDA synchronization.
- **Validation:** Packing, sampler reuse, budget and kinematic workflows pass on CPU/CUDA. General ring-closure/pucker sampling remains outside scope. Updated stack CI and the expanded CIF/FastRelax sweep are in progress.

Stack 7/10, based on #514; followed by #516. Derived from Frank DiMaio’s #503 and the fixes in #508, retaining master’s #507 efficiency work.

Requires AtomWorks [#349](https://github.com/baker-laboratory/atomworks-dev/pull/349) (shared conversion/protonation) and [#351](https://github.com/baker-laboratory/atomworks-dev/pull/351) (amide/phosphoryl completion and optional CCD reading). Pinned to `6b30ea05`; merge/release remains a landing dependency.


kierandidi:

PR #520, opened 2026-09-14
Reduce preparation and connectivity overhead for general chemistry

kierandidi:
Reduce repeated preparation and connectivity overhead while retaining scoring parameters and sampling budgets.

- **Efficiency:** Cache seeded conformers with independent copies; fuse connectivity gathers; avoid redundant int64 storage and argmin temporaries. Batch hydrogen completion for guidance and skip the unused interaction graph when bump checks are disabled.
- **Correctness:** Preserve unseeded sampling, explicit bump checks, gradients and legacy construction. Remove an unmatched profiling range.

Paired H200 measurements, PyTorch 2.13/CUDA 13.2, fixed seeds and warmups:

| Workload | Before → after | Improvement |
|---|---:|---:|
| Glycan Atom37 guidance + gradients | 33.3 → 3.12 ms | 10.7× |
| Repeated glycan / MOL2 preparation | 1.592 → 0.348 s / 0.260 → 0.030 s | 4.6× / 8.7× |
| 8,192-block connectivity gather | 8.34 → 4.42 ms | 1.9× |
| Gather temporary GPU allocation | 6.04 → 2.42 GB | 60% less |

Gather timings exclude shortest-path computation. Small-workload full FastRelax timings are broadly unchanged. Large assemblies remain unresolved: 6PWB exhausted GPU memory during rotamer scoring; 9DC2 is still running.

**Validation:** Nine representative workflows pass scoring, gradients and eight-stage FastRelax. CPU/CUDA tensor-gradient, packing and connectivity checks pass; Nsight traces cover the workloads. Latest full suite: 4,604 passed and one obsolete warning assertion, now fixed and separately tested. Latest ligand/corpus subset: 133 passed. Exact-head full validation and hosted CI remain outstanding.

Stack 11, based on #518. Profiling and review artifacts remain local.

PR #521, opened 2026-09-15
[12/12] Build poses from Atom37 tensors on one path

kierandidi:
Build a `PoseStack` from Atom37 tensors on one path, for canonical and
noncanonical chemistry alike.

`pose_stack_from_atom37(coords, res_types, chain_id, context)` needs no
structure object: residue identity comes from `res_types` indexed into the
context's canonical ordering, and chemistry the polymer connections do not
describe comes from `covalent_bonds`, so glycans, ligands and conjugated
residues build from tensors. Guidance, diffusion and search workloads that
already resolved their topology no longer touch an `AtomArray` at all.

The slot layout is data rather than code. `atom37_slot_map_for_ordering()`
builds the slot-to-canonical-atom table from the residue-type table and a
supplied layout, and leaves terminus-added atoms unmapped by reading which
atoms the termini patches add from the chemical database rather than naming
them. `canonical_form_from_atomworks` now delegates its scatter here, so the
canonical amino-acid adapter and the general path share one implementation;
the `AtomArray` entry points keep their role as derivers of these tensors.

**Validation:** Canonical protein, conjugated lysine-biotin and glycan-tree
fixtures each build through the tensor path and match the AtomArray route's
block types, connectivity, coordinates and gradients. A slot map built from a
full-atom layout leaves every terminus-added atom unmapped. Broader CPU/CUDA
runs are in progress.

Stack 12/12, based on #520. Derived from Frank DiMaio's #503.


kierandidi:

